PR #1, opened 2015-11-23
Tweak to make properties the same as rdflib.js

timbl:
This change does not align methods or class names to be the same, as they can be added as equivalents.

### DIFF
--- a/interface-spec.md
+++ b/interface-spec.md
@@ -4,10 +4,10 @@
 **Purpose:** This document proposes a uniform interface to represent RDF data in low-level JavaScript libraries.
 
 ## Design elements and principles
-- We define data interfaces to represent **triples**, **quads**, **IRIs**, **blank nodes**, **literals** and **variables**.
+- We define data interfaces to represent **triples**, **quads**, **named nodes**, **blank nodes**, **literals** and **variables**.
 - Instances of the interfaces created with different libraries should be interoperable.
 - Interfaces do _not_ specify how instances are stored in memory.
-- Interfaces mandate specific pre-defined methods such as `.equals()`.
+- Interfaces mandate specific pre-defined methods such as `.sameTerm()`.
 - Given the necessity of methods, plain objects (JSON) cannot be used.
 - Factory functions (e.g., `triple()`) or methods (e.g., `store.createTriple()`) create instances.
   - Should allow "upgrading" a plain object into a fully functional triple
@@ -18,17 +18,19 @@
 
 **Properties:**
 
-- `String .value` is refined by each interface which extends Node
+- None 
 
 **Methods:**
 
-- `.equals(Node other)` returns true if and only if the argument is a) of the same type b) has the same contents (value and, if applicable, type or language)
+- `.sameTerm(Node other)` returns true if and only if the argument is a) of the same type b) has the same contents (value and, if applicable, type or language)
 
-### IRI extends Node
+### Symbol extends Node
+
+A node named with a URI.  The node is not a URI itsef.
 
 **Properties:**
 
-- `String .value` the IRI as a string (example: `http://example.org/resource`)
+- `String .uri` the IRI naming the node as a string (example: `http://example.org/resource`)
 
 ### BlankNode extends Node
 
@@ -66,14 +68,14 @@
 
 **Properties:**
 
-- `Node .graph` the named graph, which is an IRI, a BlankNode or NodeVariable.
+- `Node .why` the provenence, which may be a named node represenitng a document on the web which has this graph, or a BlankNode or NodeVariable. Also extednable for the results of infreence.
 
 ### DataFactory
 
 **Methods:**
 
-- `.iri(String iri)` returns a new instance of IRI.
-- `.blankNode()` returns a new instance of BlankNode.
+- `.sym(String iri)` returns a new instance of IRI.
+- `.bNode()` returns a new instance of BlankNode.
 - `.literal(String value, String language, String datatype)` returns a new instance of Literal.
 - `.variable(String name)` returns a new instance of NodeVariable. This method is optional.
 - `.triple([Object])` returns a new instance of Triple. 


timbl:
The class name changes and the factory methods are not critical, but the properties are.   Named nodes have .uri not .value for their URI, and quads have .why rather than .graph, as the provenance is not a graph. , need to reserve the concept of graph for a set of triples (or quads).

RubenVerborgh:
Hi @timbl, thanks for starting this discussion. I propose we take this up in the next [weekly meeting](https://github.com/rdfjs/rdfjs.org/wiki/Representation-Task-Force#meetings).

My personal opinion: I'm not really in favor of the suggested renamings, as `.value` can be used for both literals and IRIs (whereas `.uri` cannot). A similar reasoning holds for `.equals`, which is more generic than `.sameTerm`.

But perhaps I did not fully understand

> This change does not align methods or class names to be the same, as they can be added as equivalents.

Could you clarify?

Your other point

> Named nodes have .uri not .value for their URI, and quads have .why rather than .graph, as the provenance is not a graph. , need to reserve the concept of graph for a set of triples (or quads).

actually came up last meeting. @bergos wondered whether a graph could be something else than a collection of quads with the same “fourth element”. I think that in RDF 1.1, this would actually be the case, i.e., the “fourth element” would really be `.graph`. I'm not sure to what extent provenance would be a dominant use case for this.

All of this is open for discussion of course, and we appreciate this and any other input from library developers for the task force.

elf-pavlik:
> quads have .why rather than .graph, as the provenance is not a graph. , need to reserve the concept of graph for a set of triples (or quads).

Thinking of "Content-Type: application/trig" responses which @RubenVerborgh uses as examples in http://ruben.verborgh.org/blog/2015/10/06/turtles-all-the-way-down/

```trig
<#about> {
  <#about> foaf:primaryTopic <>.
  <> a hydra:Collection, hydra:PagedCollection;
     dc:title "Amazon results for 'Peppa Pig'";
     hydra:totalItems 30862;
     hydra:itemsPerPage 16;
     hydra:firstPage </search/?q=peppa+pig&page=1>;
     hydra:nextPage </search/?q=peppa+pig&page=2>.
  </#amazon> ex:paymentCondition [
    ex:condition [ ex:minimumTotal 35.00 ];
    ex:result [ ex:shippingCost 0.00 ]
  ].
}
</items/45158567#id> a pto:Book;
  schema:name "Peppa Goes Swimming"@en.
</items/35235179#id> a pto:Toy;
  schema:name "Peppa Pig Regular Plush"@en.
</items/10268448#id> a pto:Dress;
  schema:name "Peppa Pig Long Sleeve Bicycle Gown"@en.
```

To keep provenance we would need to use both *why* and *graph*

```
{
  "subject": "http://amazon.example/search/?q=peppa",
  "predicate": "http://www.w3.org/ns/hydra/core#firstPage",
  "object": "http://amazon.example/search/?q=peppa+pig&page=1",
  "graph": "http://amazon.example/search/?q=peppa#about"
  "why": "http://amazon.example/search/?q=peppa"
}
```
?

RubenVerborgh:
But can't provenance be stores as graphs? This is what, for instance, nanopublications do ([example](http://nanopub.org/wordpress/?page_id=57)).

interface-spec.md
```diff
@@ -4,10 +4,10 @@
 **Purpose:** This document proposes a uniform interface to represent RDF data in low-level JavaScript libraries.
 
 ## Design elements and principles
-- We define data interfaces to represent **triples**, **quads**, **IRIs**, **blank nodes**, **literals** and **variables**.
+- We define data interfaces to represent **triples**, **quads**, **named nodes**, **blank nodes**, **literals** and **variables**.
```
retog: +1

retog: -1, I see no reason to differ from the terms used in the rdf 1.1 abstract syntax: “The abstract syntax has two key data structures: RDF graphs are sets of subject-predicate-object triples, where the elements may be IRIs, blank nodes, or datatyped literals.“

melvincarvalho:
What I understood above was:

- You have a triple which is a statement about something
- That triple comes into being for a reason
- The "why" is that reason
- A graph is a set of triples or quads

I could be wrong, but it seems graph values are very often the why, but need not be so.

retog:
@melvincarvalho a graph is a set of triples, a set of quads is a set of named graphs. The why of *t*  (the reason why *t* is asserted) might be the fact that a graph *g* contains *t*. The elements  of *t* (*s*, *p*,*o*) and *g* form a quad together.

The why of t might also be some triples t1 and t2  from g1 and g2 from which t can be inferred.

So to summarize:
 * A quad *q* (*s*,*p*,*o*,*g*) has a `graph` *g*
 * If *g* is justified and *q* then the triple *t*  (*s*,*p*,*o*) is justified => in this case *g* is the "why" of t
 * As written by @timbl  in the proposed patch the "why" might also be "the results of inference"

In my opinion what we should model here are graphs and the quads resulting from adding the graphname as the fourth element to the triple. While that graphname can be a justification for the triple I think this shouldn't be our concern here.

elf-pavlik:
[In gitter chat](https://gitter.im/rdfjs/publicat=565424eb77b7a53a219e44d7) @RubenVerborgh suggested that core representation, which this group works on, would only focus on quads (triples in named graphs). Than rdflib.js and other libraries could extend it with additional features, like tracking how those triples/quads ended up in the dataset eg. using *why* property.

@timbl how rdflib.js represents responses with "Content-Type: application/trig" or "Content-Type: application/ld+json", which include named graphs?

interface-spec.md
RubenVerborgh: Discussed in the 2015-11-27 meeting: We like to follow the RDF abstract syntax and keep "IRI" as a term.

interface-spec.md
```diff
@@ -4,10 +4,10 @@
...
 - Instances of the interfaces created with different libraries should be interoperable.
 - Interfaces do _not_ specify how instances are stored in memory.
-- Interfaces mandate specific pre-defined methods such as `.equals()`.
+- Interfaces mandate specific pre-defined methods such as `.sameTerm()`.
```
RubenVerborgh: Discussed in the 2015-11-27 meeting: we prefer to keep `.equals`, which checks both the type and the value.

interface-spec.md
```diff
@@ -18,17 +18,19 @@
...
 
-### IRI extends Node
+### Symbol extends Node
+
+A node named with a URI.  The node is not a URI itsef.
 
 **Properties:**
 
-- `String .value` the IRI as a string (example: `http://example.org/resource`)
+- `String .uri` the IRI naming the node as a string (example: `http://example.org/resource`)
```
RubenVerborgh: Discussed in the 2015-11-27 meeting: we prefer to keep `.value` for uniformity with literals, but we will also need to add a `.type` or similar to distinguish between IRIs/literals/blank nodes/variables.

However, this choice means that checking whether something is an IRI and reading its value requires two steps. Yet it avoids the problem where an object would have both `.value` and `.uri`.

interface-spec.md
```diff
@@ -18,17 +18,19 @@
 
 **Properties:**
 
-- `String .value` is refined by each interface which extends Node
+- None
```
RubenVerborgh: See line 33.

interface-spec.md
```diff
@@ -18,17 +18,19 @@
...
 
 **Methods:**
 
-- `.equals(Node other)` returns true if and only if the argument is a) of the same type b) has the same contents (value and, if applicable, type or language)
+- `.sameTerm(Node other)` returns true if and only if the argument is a) of the same type b) has the same contents (value and, if applicable, type or language)
 
-### IRI extends Node
+### Symbol extends Node
+
+A node named with a URI.  The node is not a URI itsef.
```
RubenVerborgh: Discussed in the 2015-11-27 meeting: RDF 1.1 seems to assume that the node _is_ the URI.

interface-spec.md
```diff
@@ -66,15 +68,15 @@
 
 **Properties:**
 
-- `Node .graph` the named graph, which is an IRI, a BlankNode or NodeVariable.
+- `Node .why` the provenence, which may be a named node represenitng a document on the web which has this graph, or a BlankNode or NodeVariable. Also extednable for the results of infreence.
```
RubenVerborgh: Discussed in the 2015-11-27 meeting: `.graph` is the more generic meaning; it is not necessarily for provenance.

However, individual implementations are free to extend objects with other properties, such as `.why`. In general, we have to be careful about the handling of such extra properties, when serializing, cloning, etc. Other examples include line number, DOM element, …

interface-spec.md
```diff
@@ -66,15 +68,15 @@
...
 
 ### DataFactory
 
 **Methods:**
 
-- `.iri(String iri)` returns a new instance of IRI.
-- `.blankNode()` returns a new instance of BlankNode.
+- `.sym(String iri)` returns a new instance of IRI.
+- `.bNode()` returns a new instance of BlankNode.
```
RubenVerborgh: Discussed in the 2015-11-27 meeting: we prefer the old names, because we want to follow the RDF spec names, which are more explanatory.

timbl: The problem with "IRI" is that the node is NOT an IRI.  It has one.   A node which is labelled by an IRI may be referred to in the RDF spec loosely as an IRI but it is not an IRI, an IRI is a string.   I am much less concerned about methods than properties, as is easy to add both methods.

RubenVerborgh: We debated about this topic during the call, especially the suggestion to call them “named nodes” rather than “IRIs”. However, [“named node” not used in any of the core RDF specifications](https://www.google.com/search?q=site:www.w3.org%2FTR%2F+%22RDF%22+%22named+node%22), so it might not be clear/intuitive.

That said, I do see the argument that an IRI is a string, and thus not a node; the confusion is especially apparent in the method signature `IRI iri(String iri)`. Maybe we should indeed call them `NamedNode` and create them with `.namedNode`. @bergos, @l00mi, @nicola?

Note that properties are as easy to add as methods in JavaScript; they have the exact same flexibility thanks to `Object.defineProperty`.

interface-spec.md
timbl: .graph is "more generic?  just a general field which can be used for anything?   It would not have to point to a graph object?   Maybe i have missed a deeper discussion

interface-spec.md
timbl: Sigh

interface-spec.md
RubenVerborgh: `.graph` would be an IRI (named node). So given the following N-Quads:
```
<a> <b> <c> <d>.
```
the object would be
```
{ subject: a, predicate: b, object: c, graph: d }
```
where `a`, `b`, `c`, `d` are IRIs (named nodes). In particular, `d` is _not_ a graph object.

interface-spec.md
RubenVerborgh: In any case, [`Symbol` is an ES6 name](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol) so we should avoid using that name in a library. While I can see the argument that _“IRI ≠ node with that IRI”_, all RDF specs use the term “IRI” in practice. It might consequently be the term that is easiest to use for developers.

The important questions are: is the difference between IRI and Node important for JavaScript libraries, how does it affect developers, and which choice is most practical?

interface-spec.md
Acubed: Isn't a Node, here, just a datum in an RDF Statement? An IRI is not a _resource_, is my understanding.

Why do we call them bNodes, if those also aren't nodes?

[RDF Interfaces](http://www.w3.org/TR/rdf-interfaces/) uses NamedNode, and has many implementations. We already thought out much of this, why the desire to re-implement from scratch?

RubenVerborgh: Regarding RDF Interfaces, the goal of this Task Force is to come up with a JavaScript-idiomatic interface. RDF Interfaces can provide the building blocks (pointers to NamedNode discussions would be welcome), but other aspects are not straightforward to reuse (e.g., strong focus on synchronicity, whereas the single-threaded nature of JavaScript requires asynchronicity to execute certain tasks in a non-blocking way).

l00mi: As it only reflects a name and not any architecture or functionality decision I don't see it too important. Following thoughts: 

1) Unfortunately is the term NamedNode not mentioned neither in [RDF1.1](http://www.w3.org/TR/rdf11-concepts/) nor in [TURTLE](http://www.w3.org/TR/turtle/). Than again because this low level API is focused mostly on experts, they should find their way around anyway.

2) I like to understand if a NamedNode simply is defined as disjoint to BlankNode (like in [RDF1.1](http://www.w3.org/TR/rdf11-concepts/))? In this case would the Literal be also a NamedNode?

On a purely practical level personally I feel that `.iri` is easier to grasp and work with in the daily business.

Acubed: @RubenVerborgh RDF Interfaces merely provides a primitive for representing RDF structures in memory; for queries it serves the same purpose that Array or String does to an asynchronous RDB query library (think SQL or IndexedDB): Representing result sets.

There's other interfaces that might be better targets for asynchronicity, e.g. [RDF API](http://www.w3.org/TR/rdf-api/). Similarly, my adaption of IndexedDB to RDF graphs, that I call [IndexedGraph](https://lists.w3.org/Archives/Public/public-rdfjs/2013Oct/0040.html), still uses RDF Interfaces for handing values in memory.

If for this purpose, the existing implementation is insufficient or has problems (and it does, hence the desire to take up work), we should be at a point where we understand and can clearly articulate why.

timbl:
I think I have not helped by using the work "provenance".   Let's not have a fifth slot.  Here are the main ways that .why slot is used in the code I use. 

- A document (a named node with the .uri being the URI of the document) as a shorthand for "The graph which we got from parsing the document at u". (Or more generally, the graph corresponding to that document, whether or not it has just been parsed.)   This is set by a parser, and a serializer can serialize only the quads matching that document, or optionally serialize all the quads in the store.  This is used when you have loaded a bunch of documents, but want to make sure that for example an Access Control triple was in the proper ACL file not some random other data.  The API needs to allow easily working with the graph from one document or the union graph.

- An graph object for triples which have been parsed between {} in n3, and can occur in S P or O positions in other quads in the store.  
In this case you don't need to store the fact that the original n3 indirectly contains the triple.  You just record that it contains a nested graph at the top level and then the triple in in that graph, that gives you all you need.

- A  bnode used to optionally store the HTTP response headers as triples, to allow code to operate on those, e.g. check method supported, etc

- (in cwm) when inference rules have produced triples, optionally an object pointing to the rule, the parameters and the incoming data which fired the rule.  This is obviously an extension for advanced stuff with proofs.   But it did just use the same .why slot, not any extra slot.

If we were designing from scratch, the property ".graph" would be fine for the first three cases except it would break existing code.  Does using ".graph" seem more intellectually satisfying than ".why"?  Not really. It works well for the second case, maybe the third.  "Why" makes english sense for all the cases except for the second. 

Trig is a funny format in that it is designed to dump a quadstore, but it is not itself RDF on n3.  It initialized the store to have the same .why values as from when the store was  dumped.  (If you use an n3 vocabulary to express the content of the store, then you have two different operations, one to restore the store from the dump, the other to load the n3 as the normal way, in which case you can explore the equivalent of Trig statements as top-level triples.)

In general by the way I applaud this effort to produce a common standard and like everyone I guess I am weighing the value of achieving consensus vs changing code.

westurner:
Named Graphs / Provenance 
* | Wikipedia: https://en.wikipedia.org/wiki/Named_graph
* | Docs: http://patterns.dataincubator.org/book/named-graphs.html

N3 / TriG / JSON-LD

* N3 Standard: http://www.w3.org/TeamSubmission/n3/
* TriG Standard: http://www.w3.org/TR/trig/
* JSON-LD Standard: http://www.w3.org/TR/json-ld/#named-graphs

Named Graph standards gaps:

- [[ ]] 'port' ``=>`` implies to TriG
- [[ ]] 'port' named graph support to N3

...

* PROV Standard: http://www.w3.org/TR/prov-o/
  * https://pypi.python.org/pypi/prov
  * https://wrdrd.com/docs/consulting/knowledge-engineering#prov

RubenVerborgh:
Will be resolved with other issues.

Issue #6, opened 2015-12-03
Specification of .toString should be delegated to implementers


bergos:
It is not this library's task to specify the behavior of `.toString`, as specific implementations might like to add different behavior.

RubenVerborgh:
See #7 for a related method that we will specify.

RubenVerborgh:
We should mention that we don't specify `.toString`, and list the related functionality we do have (`.value`, `toCanonical()`).

RubenVerborgh:
See #19.

Issue #5, opened 2015-12-03
.value should not have prefixes for variables and blank nodes


bergos:

RubenVerborgh:
Agreed during the call.

RubenVerborgh:
See #19.

dlongley:
Not sure where this discussion should be taking place... here's a link to my recent comments:

https://github.com/rdfjs/representation-task-force/pull/19#issuecomment-176852702

RubenVerborgh:
Resolved with #36.

PR #19, opened 2015-12-13
removed leading _: and ?, list .toString() as not specified

bergos:
- .value should not have prefixes for variables and blank nodes [#5](https://github.com/rdfjs/representation-task-force/issues/5)
- Specification of .toString should be delegated to implementers [#6](https://github.com/rdfjs/representation-task-force/issues/6)

### DIFF
--- a/interface-spec.md
+++ b/interface-spec.md
@@ -38,9 +38,7 @@ TODO: to what extent should we use typed signatures (`.equals(Node other)`) vers
 
 **Properties:**
 
-- `String .value` blank node name as a string (example: `_:blank3`)
-
-TODO: Does the value always start with an underscore?
+- `String .value` blank node name as a string without leading `_:` (example: `blank3`)
 
 ### Literal extends Node
 
@@ -58,9 +56,7 @@ TODO: Why not just Variable?
 
 **Properties:**
 
-- `String .value` the name of the variable (example: `?a`)
-
-TODO: Does the value always start with a question mark?
+- `String .value` the name of the variable without leading `?` (example: `a`)
 
 ### Triple
 
--- a/interface-spec.md
+++ b/interface-spec.md
@@ -28,6 +28,11 @@ TODO: read/write or read-only?
 
 TODO: to what extent should we use typed signatures (`.equals(Node other)`) versus actual JavaScript signatures (`.equals(other)`). The benefit of typed signatures is that you see the type inline; the drawback is that it is more specific than JavaScript itself. For ease of use, JavaScript might be preferred, specifying types in the explanation (or jsdoc-style).
 
+**Methods not specified:**
+
+- `.toString()` is not specified, therefore the results may vary for each implementation.
+The property `.value` and method `.toCanonical()` should be used to access the value of a term.
+
 ### IRI extends Node
 
 **Properties:**
--- a/interface-spec.md
+++ b/interface-spec.md
@@ -1,8 +1,4 @@
-# Interface Specification: RDF Representation
-
-## Mission Statement
-This document provides a specification of a low level interface definition representing RDF data independent of a serialized format in a JavaScript environment. The task force which defines this interface was formed by RDF JavaScript library developers with the wish to make existing and future libraries interoperable. This definition strives to provides the minimal necessary interface necessary to enable interoperability of libraries such as serializers, parsers and higher level accessors and manipulators.
-
+# Interface Specification
 
 **Status:** This document is a _draft_.<br>
 **Purpose:** This document proposes a uniform interface to represent RDF data in low-level JavaScript libraries.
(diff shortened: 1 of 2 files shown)


interface-spec.md
```diff
@@ -28,6 +28,11 @@ TODO: read/write or read-only?
 
 TODO: to what extent should we use typed signatures (`.equals(Node other)`) versus actual JavaScript signatures (`.equals(other)`). The benefit of typed signatures is that you see the type inline; the drawback is that it is more specific than JavaScript itself. For ease of use, JavaScript might be preferred, specifying types in the explanation (or jsdoc-style).
 
+**Methods not specified:**
+
+- `.toString()` is not specified, therefore the results may vary for each implementation.
+The property `.value` and method `.toCanonical()` should be used to access the value of a term.
+
```
RubenVerborgh: I'm having second thoughts about this. A “methods not specified” section makes things unnecessarily complicated.

I think `.toString()` can just be a note with `.toCanonical()`. Also, such a note should aim to explain the decision: _`.toCanonical()` is the only textual representation guaranteed to be consistent across implementations; other methods like `.toString()` might contain implementation-specific information._

interface-spec.md
```diff
@@ -38,9 +43,7 @@ TODO: to what extent should we use typed signatures (`.equals(Node other)`) vers
 
 **Properties:**
 
-- `String .value` blank node name as a string (example: `_:blank3`)
-
-TODO: Does the value always start with an underscore?
+- `String .value` blank node name as a string without leading `_:` (example: `blank3`)
```
RubenVerborgh: For consistency with the description for variable:

the name of the blank node without leading `_:`

interface-spec.md
bergos: I agree on the "methods not specified". How many of them will we have? I expect it will be only `.toString()`. But people will search for it. Beside your proposal for the `.toCanonical()` I would add the `.toString()` method to the list of methods with a text like: "not specified, see `.toCanonical()`". If it's only on the `.toCanonical()` text, it's a little bit hidden. We should do this only for the special case of `.toString()`!

interface-spec.md
bergos: +1

interface-spec.md
RubenVerborgh: Or people can simply search for `toString`, which would bring them to `.toCanonical()` anyway then.

interface-spec.md
dlongley: -1

I don't see the value of removing `_:` from the value -- that is actually part of the identifier for the blank node. Furthermore, blank node identifiers are not part of the RDF abstract syntax -- the presumption that they start with `_:` is only based on the fact that nearly all concrete syntaxes use this notation. While unlikely, it's *possible* that another syntax would not use that form.

If the actual value of a blank node identifier is not going to be found via the `.value` property, then I think there should be a very good argument for taking that approach. IMO, this approach would be unexpected.

RubenVerborgh: > presumption that they start with _: is only based on the fact that nearly all concrete syntaxes use this notation. While unlikely, it's possible that another syntax would not use that form.

Which is why `_:` should  not be there, because it is syntax-specific?

ericprud: * Dave Longley <notifications@github.com> [2015-12-18 06:26-0800]
> > @@ -38,9 +43,7 @@ TODO: to what extent should we use typed signatures (`.equals(Node other)`) vers
> >  
> >  **Properties:**
> >  
> > -- `String .value` blank node name as a string (example: `_:blank3`)
> > -
> > -TODO: Does the value always start with an underscore?
> > +- `String .value` blank node name as a string without leading `_:` (example: `blank3`)
> 
> -1
> 
> I don't see the value of removing `_:` from the value -- that is actually part of the identifier for the blank node. Furthermore, blank node identifiers are not part of the RDF abstract syntax -- the presumption that they start with `_:` is only based on the fact that nearly all concrete syntaxes use this notation. While unlikely, it's *possible* that another syntax would not use that form.

So doesn't that make the _: a property of the syntax?

It's true that at present, there's no promise that one can parse that reference the same bnode label and have them unify on that label.
It is quite reasonable that behind closed doors one would want to do that in order to leverage internal identifiers.

  RDF/XML: <rdf:Description rdf:nodeId="proteinProductTP53"><rdfs:label>P53 product</rdfs:label></rdf:Description>
+ Turtle:  _:proteinProductTP53 a :GeneProduct .
= _:proteinProductTP53 a :GeneProduct ; rdfs:label "P53 product" .

That's a bit heretical but quite practical. It also illustrates that the _: is just Turtle's syntax for flagging a bnode.


> If the actual value of a blank node identifier is not going to be found via the `.value` property, then I think there should be a very good argument for taking that approach. IMO, this approach would be unexpected.
> 
> ---
> Reply to this email directly or view it on GitHub:
> https://github.com/rdfjs/representation-task-force/pull/19/files#r48028993

-- 
-ericP

office: +1.617.599.3509
mobile: +33.6.80.80.35.59

(eric@w3.org)
Feel free to forward this message to any list for any purpose other than
email address distribution.

There are subtle nuances encoded in font variation and clever layout
which can only be seen by printing this message on high-clay paper.

dlongley: > Which is why _: should not be there, because it is syntax-specific?

The entire identifier is syntax-specific. That's why it's odd to break it apart.

interface-spec.md
bergos: +1

interface-spec.md
dlongley: > So doesn't that make the _: a property of the syntax?

The whole identifier is a property of the syntax.

> [combination of two different syntaxes example]
> That's a bit heretical but quite practical. It also illustrates that the _: is just Turtle's syntax for flagging a bnode.

Well, I think that example is bit presumptuous, but I do understand the practicality of it. I do think it's quite useful to preserve blank node identifiers (eg: for representing canonicalized data sets, for example) even if they aren't in the abstract data model. But there's a consistency issue based on the current wording.

You can make a similar example:

```
  RDF/XML: <rdf:Description rdf:nodeId="_:proteinProductTP53"><rdfs:label>P53 product</rdfs:label></rdf:Description>
+ Turtle:  _:proteinProductTP53 a :GeneProduct .
= _:proteinProductTP53 a :GeneProduct ; rdfs:label "P53 product" .
```

And I believe it's just as "correct" if we're consistently following the wording given. This would draw the opposite conclusion, that `_:` is not **just Turtle's** syntax for flagging a bnode. We're applying `_:` to all source syntaxes.

**Just imagine if we parse `<... rdf:nodeId="_:b">` and <... rdf:nodeId="b">` from the same document. Are we going to have a collision that mangles the dataset using the rules laid out above?**

In Turtle, blank nodes are described as the combination of `_:` and a "blank node label". And it is true that two blank nodes having the same label are in fact to be considered the same node. That does seem to make them *identifiers* within that syntax. However, that doesn't mean that this rule extends across other syntaxes. RDF/XML has no concept of "blank node labels", only "blank node identifiers".

So, the real problem here is that we're taking syntax-specific rules and leaking them into our representation of the data model. However, our representation can't preserve the graph relationships without blank node identifiers. This means we could either generate fresh ones -- or we can try and preserve the ones from the concrete syntax from which we loaded the dataset. Again, I do actually think it's important to do so because it makes canonicalizing data sets possible.

So, all of that is to say that I think perhaps my issue can be resolved by reconsidering the wording "blank node name as a string without leading `_:`". This presumes too much about the source concrete syntax. We need some other wording that is more general or perhaps we could address it via a parenthetical:

"blank node identifier as a string (note: for clarity, if the identifier was parsed from a concrete syntax that prefixes blank node identifiers with `_:`, this prefix is removed).

This would make it clear that when you parse `rdf:nodeId` from RDF/XML, if the value is taken *as is*, even if `_:` is a prefix. If you parse from Turtle, `_:` is removed. This should avoid mangling data.

RubenVerborgh: Thanks a lot for the detailed comment. I will indeed adjust the wording there.

bergos: Any proposals for the wording? Should we replace blank node name with blank node identifier?

bergos:
Accidentally @l00mi merge the PRs into my repository. Anyway, we wanted to skip most of the changes. Can we finish the discussion, than I will create a new PR?

dlongley:
@bergos, if you're asking about:

> `String .value` blank node name as a string without leading `_:` (example: `blank3`)"

The point I was making was that different syntaxes do different things. For example, RDF/XML doesn't prefix `_:` ... however, if someone did that on their own in an RDF/XML doc, then it should be preserved because otherwise you could have data collision. Turtle, on the other hand, does prefix `_:` and, therefore, those should be removed.

So it's source syntax specific as to whether or not the `_:` should appear once the data has been parsed.

bergos:
@dlongley OK, now I understand. So can we write something like:

blank node name as a string without any serialization specific prefixes, like _: for Turtle and N-Triples.

dlongley:
> blank node name as a string without any serialization specific prefixes, like _: for Turtle and N-Triples.

This is better, but not quite precise enough. We don't want people to think that one way to produce this representation when parsing a syntax is to go down a list of blank node prefixes for known syntaxes and remove any matching prefix from a blank node name if found. Instead, we want people who parse data from a particular syntax to ensure that they remove the blank node prefix used *by that syntax*, if any. That might seem a bit pedantic, but it's best to avoid confusion. 

Perhaps:

> "blank node name as a string, without any serialization specific prefixes, e.g. if the data was sourced from Turtle, remove `_:`, if it was sourced from RDF/XML, do not change the blank node name."

Or something to that affect -- that shows two different cases.

PR #23, opened 2016-01-19
added DefaultGraph interface and .defaultGraph factory method

bergos:


interface-spec.md
```diff
@@ -99,6 +109,7 @@ TODO: Do we need to define a different interface, or is a quad simply a triple w
 - `.blankNode()` returns a new instance of BlankNode.
 - `.literal(String value, String language, String datatype)` returns a new instance of Literal.
 - `.variable(String name)` returns a new instance of Variable. This method is optional.
+- `.defaultGraph()` returns an instance of DefaultGraph. This method should return a singleton.
```
RubenVerborgh: "This method should return a singleton." => "This method always returns the same constant."

interface-spec.md
```diff
@@ -71,6 +71,16 @@ TODO: What if the literal has no language? Does it always have a datatype?
...
 
+### DefaultGraph extends Term
+
+An instance of `DefaultGraph` represents the default graph.
+It's only allowed to assign a `DefaultGraph` to the `.graph` property of a `Quad`. 
+
+**Properties:**
+
+- `String .termType` contains the constant `"defaultGraph"`.
+- `String .value` contains an empty string as constant value.
```
RubenVerborgh: Perhaps that `.equals` is true iff the other thing's `.termType` is `"defaultGraph"`?

RubenVerborgh:
:+1: with comments

interface-spec.md
bergos: If multiple libraries are used it can't be the same constant.
Am 22.01.2016 10:11 vorm. schrieb Ruben Verborgh <notifications@github.com>:In interface-spec.md:
> @@ -99,6 +109,7 @@ TODO: Do we need to define a different interface, or is a quad simply a triple w
>  - `.blankNode()` returns a new instance of BlankNode.
>  - `.literal(String value, String language, String datatype)` returns a new instance of Literal.
>  - `.variable(String name)` returns a new instance of Variable. This method is optional.
> +- `.defaultGraph()` returns an instance of DefaultGraph. This method should return a singleton.

"This method should return a singleton." => "This method always returns the same constant."

RubenVerborgh: Right, but within a library, it must be a constant?

bergos: Yes, but with constant someone could expect === can be used to compare it. Maybe add "an implementation must use the same constant".
Am 22.01.2016 10:28 vorm. schrieb Ruben Verborgh <notifications@github.com>:In interface-spec.md:
> @@ -99,6 +109,7 @@ TODO: Do we need to define a different interface, or is a quad simply a triple w
>  - `.blankNode()` returns a new instance of BlankNode.
>  - `.literal(String value, String language, String datatype)` returns a new instance of Literal.
>  - `.variable(String name)` returns a new instance of Variable. This method is optional.
> +- `.defaultGraph()` returns an instance of DefaultGraph. This method should return a singleton.

Right, but within a library, it must be a constant?

RubenVerborgh: You're right about `===`. Perhaps remove the singleton remark altogether, as this can also set an expectation about `===`.

interface-spec.md
bergos: This could open the door to use `.value` for other stuff. Let's handle it like the other TermTypes.

RubenVerborgh: Okay, the treatment of `equals` does indeed not need to be mentioned.

pieroit:
Am I losing something important, but why don't use just a `undefined` for a missing forth term?

bergos:
We want to avoid the check for `undefined` before doing this `triple.graph.value === otherTriple.graph.value` or this `triple.graph.termType === otherTriple.graph.termType`.

bergos:
I changed the singleton stuff, so this PR should be ready to merge.

l00mi:
merged and reshuffled to correct order with https://github.com/rdfjs/representation-task-force/commit/f06307652fa1c127cd84d95efb14f06c9a0d356d

PR #24, opened 2016-01-21
added definition for canonical representation

bergos:


RubenVerborgh:
:+1: with the comments processed

bergos:
I changed the "non-default" text.

@RubenVerborgh Do you think we should include the text about the whitespace or was it just a comment for the implementers? Sure, this is important because many people would implement it with a space before the final dot, but I would like to avoid text duplication as much as possible.

RubenVerborgh:
It's just that if you mention

> **The whitespace following subject, predicate, object and graph MUST be a single space, (U+0020).**

you should also mention the other whitespace thing.

(I'm fine with mentioning neither.)

dlongley:
> So no whitespace before the final dot.

My understanding is that there should be exactly one space following the object or graph position, which would be the space right before the final dot (meaning, there is exactly one space before the dot).

RubenVerborgh:
Oh yes, of course. This can be clarified by saying:

> subject, predicate, object (and if present, graph) should be followed by exactly one space.

dlongley:
Triples or quads are separated by a dot `.` and a new line `LF`; should we also say that a `LF` is *always* present in canonical form, even at the end of the document? (This is how it is presently done in the normalization algorithm).

RubenVerborgh:
That seems very important indeed!

bergos:
You are right about the space before the dot.

The spec says: `This sequence is terminated by a '.' and a new line (optional at the end of a document)` in the introduction and `All other locations that allow whitespace MUST be empty.` in the canonical form section.

If there would be no definition in the spec, I would vote for `LF` at the end of the document. But there is a definition, so let's follow the spec.

RubenVerborgh:
> But there is a definition, so let's follow the spec.

Alright, but the definition here does not state whether or not there should be an LF, since we are technically "at the end". So do we require an LF or not?

dlongley:
Do we have any data on how others have interpreted the spec?

RubenVerborgh:
They usually had a different context: representing actual documents of triples. There, the only question is whether or not to end the file with an LF, all the rest is fixed.

Here, we aim to use this for single-triple representations instead, so comparisons might not be representative.

bergos:
@RubenVerborgh you are right, we just handle terms and triples. The LF would be optional. So do we skip it? Should we write this in our spec?

bergos:
Changes relalated to the LF stuff should be handled in a new PR.

blake-regalia:
EOL is platform-dependent. In the N-Triples spec, the EOL production is `[#xD#xA]` which is `'\r\n'` (carriage-return followed by newline). Although this is standard for Windows EOL, this is not the case for *nix and most string uses in javascript. For these reasons (among others), I propose that `Quad#toCanonical` does not produce a terminal EOL.

PR #25, opened 2016-01-21
added sentence about additional properties and wiki link

bergos:


RubenVerborgh:
What is the intended functionality of the list of properties? Is it necessary/scalable to keep a list of them, or is this something we (should) leave up to the implementers?

pieroit:
Keeping a list of extra properties, at least in the beginning, could be useful to understand how the spec is used and what could be changed.
:+1:

bergos:
We don't want to spec additional properties, but also it would be nice if the community reuses property names for the same use cases. The wiki of this task force would be a good place to maintain a list. It's up to the implementers how they use the properties, but if more implementers agree on some properties, that could be used for an additional spec in the future.

PR #34, opened 2016-02-04
moved TODOs to issues

bergos:
We discusses already many TODOs. I created issues for all others.


RubenVerborgh:
:+1: verified all TODOs; they either were fixed, were already issues, or got a new issue

PR #37, opened 2016-02-05
Describe constructor of Quad/Triple

RubenVerborgh:
Right now, the argument is an unknown object. We should at least provide one way for the first draft.
Suggestion: `Quad(Term, Term, Term, Term)`


RubenVerborgh:
@bergos Turned this issue into a pull request.

Issue #38, opened 2016-02-05
Quad extends Triple


rubensworks:
The [interface spec figure](https://github.com/rdfjs/representation-task-force/blob/v0.1/interface-spec.md#data-interfaces) shows that Quad extends Triple.
However, In the description of Triple, it is said that Triple is an alias of Quad.
Wouldn't it therefore make more sense to let Triple extend from Quad?
Or am I missing something?

RubenVerborgh:
The spec text is actually right here: we decided to make Triple an alias of Quad, so that there is no difference and that they can transition seamlessly.

It is thus the figure that should be adjusted.

pieroit:
If we delete the extension arrow then we need a way to show the `alias` relation. I see two options:
- draw a single class with two names `Triple | Quad`. The problem is then that `.graph` has a predefined value only for Triple.
- another type of arrow? Can't find the UML notation for an alias.

pieroit:
Today we discussed a little about this.
A proposal is to drop the ["Quad extends Triple"](https://github.com/rdfjs/representation-task-force/blob/master/interface-spec.md#quad-extends-triple) sentence in the spec as well as the "Triple" box in the [UML](https://github.com/rdfjs/representation-task-force/blob/master/interface-spec.md#data-interfaces), because we all agree on having Triple as an alias of Quad (no extension).

The factory methods signatures should be:
`.triple(Term, Term, Term): Quad`
`.quad(Term, Term, Term, Term): Quad`

Note that both return Quad (there is no Triple object, just a Quad with`.graph = DefaultGraph`) and Triple does not take the fourth argument, which is quite misleading.

Hope this is fine @bergos @elf-pavlik @dmitrizagidulin @RubenVerborgh

PR #51, opened 2016-03-02
Triple factory shouldn't accept graph name

elf-pavlik:
> Triple is an alias of Quad. Triples always have .graph set to DefaultGraph
&mdash; https://github.com/rdfjs/representation-task-force/blob/master/interface-spec.md#triple

Since Triple has always .graph set to DefaultGraph, it doesn't make sense to allow graph as an optional parameter in factory.

### DIFF
--- a/interface-spec.md
+++ b/interface-spec.md
@@ -115,5 +115,5 @@ Triples always have `.graph` set to DefaultGraph.
 - `Literal .literal(String value, String language, String datatype)` returns a new instance of Literal.
 - `Variable .variable(String name)` returns a new instance of Variable. This method is optional.
 - `DefaultGraph .defaultGraph()` returns an instance of DefaultGraph.
-- `Triple .triple(Term subject, Term predicate, Term object, [Term graph])` returns a new instance of Triple.
+- `Triple .triple(Term subject, Term predicate, Term object)` returns a new instance of Triple.
 - `Quad .quad(Term subject, Term predicate, Term object, [Term graph])` returns a new instance of Quad.


pieroit:
+1

RubenVerborgh:
Triple and Quad _are_ aliases, both with 4 parameters; so the above sentence in the spec is actually contradictory. We have to decide one way or another, but both indeed can't be true.

bergos:
We have "Quad extends Triple" in the spec and the Quad contains all properties and methods. That doesn't make sense. We should define Triple and Quad like this "Quad | Triple" or another sign that could be used as alias. And than all function signatures should be the same.

elf-pavlik:
Maybe Triple extends Quad - by restricting it's .graph to always return DefaultGraph?

Factories don't for Quad and Triple don't act as alias and can have different signatures. If one can't set .graph on a triple to anything else than DefaultGraph, I don't see sense in making it parameter in factory.

bergos:
I think already about the implementation: I expect no one would implement it with inheritance. The classes will use the same code and share the same constructor with an optional fourth parameter. The same for the factory method. So at the end it's about whether the spec looks a little bit nicer, but the code doesn't match exactly or the spec allows a fourth parameter for triples. I think we had that already in mind, when we defined the interface like it is now.

elf-pavlik:
maybe triple factory could just wrap quad factory and ensure setting .graph to DefaultGraph?

```js
function triple (subject, predicate, object) {
  return this.quad(subject, predicate, object, this.defaultGraph())
}
```

pieroit:
Posted in #38, reposting here

Today we discussed a little about this.
A proposal is to drop the ["Quad extends Triple"](https://github.com/rdfjs/representation-task-force/blob/master/interface-spec.md#quad-extends-triple) sentence in the spec as well as the "Triple" box in the [UML](https://github.com/rdfjs/representation-task-force/blob/master/interface-spec.md#data-interfaces), because we all agree on having Triple as an alias of Quad (no extension).

The factory methods signatures should be:
`.triple(Term, Term, Term): Quad`
`.quad(Term, Term, Term, Term): Quad`

Note that both return a `Quad` (there is no Triple object, just a Quad with`.graph = DefaultGraph`) and `.triple()` does not take the fourth argument, which is quite misleading.

elf-pavlik:
resolved in #58

PR #53, opened 2016-03-14
.equals uses .toCannonical() and no uniform .value property

elf-pavlik:
this also makes property names matching parameter names in factory functions

### DIFF
--- a/interface-spec.md
+++ b/interface-spec.md
@@ -33,7 +33,7 @@ Abstract interface.
 
 **Methods:**
 
-- `boolean .equals(Term other)` returns true if and only if the argument is a) of the same type b) has the same contents (value and, if applicable, type or language)
+- `boolean .equals(Term other)` returns true if and only if the strings returned by .toCanonical() are equal
 - `String .toCanonical()` returns a canonical string representation of the term.
   For IRIs, BlankNodes and Literals the [N-Triples canonical form](https://www.w3.org/TR/n-triples/#canonical-ntriples) must be used.
   Variables must return the variable name prefixed with a question mark (example: `?a`).
@@ -43,14 +43,14 @@ Abstract interface.
 **Properties:**
 
 - `String .termType` contains the constant `"iri"`.
-- `String .value` the IRI as a string (example: `http://example.org/resource`)
+- `String .iri` the IRI as a string (example: `http://example.org/resource`)
 
 ### BlankNode extends Term
 
 **Properties:**
 
 - `String .termType` contains the constant `"bnode"`.
-- `String .value` blank node name as a string, without any serialization specific prefixes, e.g. when parsing, if the data was sourced from Turtle, remove _:, if it was sourced from RDF/XML, do not change the blank node name (example: `blank3`)
+- `String .id` blank node identifier as a string, without any serialization specific prefixes, e.g. when parsing, if the data was sourced from Turtle, remove _:, if it was sourced from RDF/XML, do not change the blank node name (example: `blank3`)
 
 ### Literal extends Term
 
@@ -69,7 +69,7 @@ Otherwise, if no datatype is explicitly specified, the datatype IRI is `http://w
 **Properties:**
 
 - `String .termType` contains the constant `"variable"`.
-- `String .value` the name of the variable without leading `?` (example: `a`)
+- `String .name` the name of the variable without leading `?` (example: `a`)
 
 ### DefaultGraph extends Term
 
@@ -79,7 +79,6 @@ It's only allowed to assign a `DefaultGraph` to the `.graph` property of a `Quad
 **Properties:**
 
 - `String .termType` contains the constant `"defaultGraph"`.
-- `String .value` contains an empty string as constant value.
 
 ### Triple
 Triple is an alias of Quad.


RubenVerborgh:
This is potentially much more expensive, given that `.toCanonical` might need to perform a computation, whereas `.value` equality can even come down to simple pointer equality. I'm reluctant to see this merged.

RubenVerborgh:
Also, this means that `.value` and `.iri` will randomly give `undefined`, whereas it used to be defined for all nodes. I think that such important deviations from the agreed-upon interface need to be extremely well-motivated, otherwise we risk reinventing ourselves constantly, and hence stagnating.

elf-pavlik:
> This is potentially much more expensive, given that .toCanonical might need to perform a computation, whereas .value equality can even come down to simple pointer equality.

does the uniform .value property has some other intended use than basing .equals() on it?

> Also, this means that .value and .iri will randomly give undefined, whereas it used to be defined for all nodes.

in what cases one would try to get .value from term other than Literal and .iri from term other than IRI (NamedNode) ?

RubenVerborgh:
> does the uniform .value property has some other intended use than basing .equals() on it?

Yes, getting the value of things, for instance for display.

> in what cases one would try to get .value from term other than Literal and .iri from term other than IRI (NamedNode) ?

In case you don't know what kind of term you are dealing with.

elf-pavlik:
I hope that I don't go through gh comments as wanting to argue against use of uniform .value property. I just want to understand better motivation for it and help with documenting it for other people who would also like to understand what motivated this choice.

> >  does the uniform .value property has some other intended use than basing .equals() on it?

> Yes, getting the value of things, for instance for display.

I struggle to imagine what kind of display you mean, especially that we discuss (Low Level)  Interface Specification


> > in what cases one would try to get .value from term other than Literal and .iri from term other than IRI (NamedNode) ?

> In case you don't know what kind of term you are dealing with.

in what example scenario this can happen, maybe one of gathered on https://github.com/rdfjs/representation-task-force/wiki/Use-Cases

RubenVerborgh:
> I just want to understand better motivation for it

Simplicity and uniformity. In object-oriented programming, we always try to reuse as much of the base class as possible. With a `.value` property, all derivatives of `Term` provide a way to see their main defining property. This is much simpler than having to remember property names for each different instance, not to mention the risk of getting `undefined` and trying to call functions on `undefined`. In our proposal, you can trust that `.value` is _always_ a string, so you can safely call string functions on it.

> I struggle to imagine what kind of display you mean, especially that we discuss (Low Level) Interface Specification

If you're writing a complex query engine, `console.log(triple.object.value)` versus `console.log(triple.object.iri || triple.object.id || triple.object.name || triple.object.value)` makes things easier, and then I'm not even talking about chimera objects that would have multiple of these properties defined.

> > In case you don't know what kind of term you are dealing with.
> in what example scenario this can happen

`triple.object`

-----

I appreciate your efforts to understand and document, thanks for this. However, as far as changing things that have been decided in an earlier phase is concerned, I'm very careful. As a group, we together decided on `.value` and have send this out on the mailing list for comments. Changing things now would mean none of the agreements we made so far are fixed, and this would endanger any progress.
While I'm open to change clear mistakes that we might have made, this decision is also one of interface and aesthetics, i.e., things we just have to agree upon and move on. So unless we made a serious mistake with `.value`, I'm highly reluctant to change it. Not because it is necessarily the _best_ option, but because it is a sufficient option we all agreed on in the past.

RubenVerborgh:
…but by any means, keep the discussion going :smile:

elf-pavlik:
> If you're writing a complex query engine, console.log(triple.object.value) versus console.log(triple.object.iri || triple.object.id || triple.object.name || triple.object.value) makes things easier

For debugging `console.log(triple.object.toCannonical()` might work just fine ...

If people would like to extend interface to allow term to have both .iri and .value, maybe on low level something bit more verbose like .getMainProperty() could work for uniform access, especially if .toCannonical() can't serve this purpose.

elf-pavlik:
Also keeping in mind [Principle of Least Surprise](https://en.wikipedia.org/wiki/Principle_of_least_astonishment). What people writing the code might expect from
`variable.value` ?
Tim made similar point in the chatroom the other day: https://gitter.im/rdfjs/public?at=56d6264806ba9a282a285667

something like .getMainProperty() could return variable.name and still give uniformity.

RubenVerborgh:
Agree on least surprise in the case of variable, however, consistency might outweigh the drawback.

I could not ever bring myself to ever type `.getMainProperty()` :wink:

elf-pavlik:
>>> In case you don't know what kind of term you are dealing with.

>> in what example scenario this can happen

> triple.object

could you past some links to your code where you access `triple.object.value` and it doesn't matter for you if you get an IRI or just textual value part of a literal?

similarly in what cases it doesn't matter when accessing `triple.subject.value` if you get an IRI or a blank node identifier but without _: prefix?

RubenVerborgh:
@elf-pavlik Honestly, I don't have such examples ready myself. That's because the N3.js library, which I use, is entirely based on canonical representations (there are no others). So that would always be my preference. However, I don't intend to call `.toCanonical` everytime.

elf-pavlik:
@bergos maybe you could provide from rdf-ext environment some examples of using value no matter if you get IRI, blank node identifier without _: or textual value of literal? https://github.com/rdfjs/representation-task-force/pull/53#issuecomment-196948962


> getting the value of things, for instance for display.

> If you're writing a complex query engine, console.log(triple.object.value) versus console.log(triple.object.iri || triple.object.id || triple.object.name || triple.object.value) makes things easier

> I could not ever bring myself to ever type .getMainProperty() :wink:

how about `.label()` similar as in *rdfs:label* ?

gkellogg:
How you deal, for example, that http://example.com/./ex is not canonical, but that it does not equal http://example.com/ex. Same with "1" as integer vs "01"? It seems that equal() is good of value equality, not term equality. 

It is reasonable for a given store to not support term equality, but it should be possible.

RubenVerborgh:
@gkellogg This is an important remark, as it indicates that we might have chosen the wrong name for `.toCanonical()`. We were not talking about canonical URIs or canonical values, but rather about a one-to-one mapping that uniquely translates a term into a string and vice-versa. (So URIs and literals with the same textual value are mapped to a different identifier.)

PR #55, opened 2016-03-18
added initial draft for stream interfaces

bergos:


interface-spec.md
```diff
@@ -117,3 +117,56 @@ Triples always have `.graph` set to DefaultGraph.
...
+  The quad is forwarded to the event listener.
+
+### Source
+
+- `Stream .read([Term|RegExp subject], [Term|RegExp predicate], [Term|RegExp object], [Term|RegExp graph])`
+  Returns a stream that processes all quads matching the pattern.
+
+### Sink
+
+- `undefined .write(Stream stream)`
```
bergos: The `.write` method could be confusing. It's more like a reverse `.pipe`, but that method name should be reserved, because some classes may implemented full featured streams. Another proposed method name was `.consume`.

It's the same for `.read`. Reading triples from a source could be also described as writing all triples from the source to a stream. `.filter` could be used, but would be confusing if the filter feature isn't used.

interface-spec.md
```diff
@@ -117,3 +117,56 @@ Triples always have `.graph` set to DefaultGraph.
...
+  This event fires if any error occurs.
+  The error message is forwarded to the event listener.
+
+- `data`
+  This event is emitted for every quad that can be read from the stream.
+  The quad is forwarded to the event listener.
+
+### Source
+
+- `Stream .read([Term|RegExp subject], [Term|RegExp predicate], [Term|RegExp object], [Term|RegExp graph])`
```
bergos: See `.write` method of `Sink`

RubenVerborgh: I think this should become something like `search`. I expect `read` to be an operation without parameters, like in `Stream` itself.

elf-pavlik:
Store extends Source, Sink

interface-spec.md
bergos: In the last meeting we collected the following alternative methods names. What do you think about them? Other proposals?

    .search
    .filter // may suggest accepting filter function
    .match // used by rdf-interfaces

interface-spec.md
bergos: In the last meeting we collected the following alternative methods names. What do you think about them? Other proposals?

    .accept
    .consume
    .input
    .import
    .append
    .readFrom
    .pull(From)
    .reversePipe

interface-spec.md
elf-pavlik: Store has `.removeMatches()` with the same signature, so `.match()` or even `.getMaches()` looks consistent.

interface-spec.md
RubenVerborgh: +1 for `import`/`readFrom`

-1 for `pull` (without `From`)/`reversePipe`

interface-spec.md
```diff
@@ -117,3 +117,56 @@ Triples always have `.graph` set to DefaultGraph.
...
+
+**Methods:**
+
+- `Quad .read()`
+  This method pulls a quad out of the internal buffer and returns it.
+  If there is no quad available, then it will return null. 
+
+**Events:**
+
+- `readable`
```
pieroit: what about `start`?

interface-spec.md
bergos: +1 `.consume / .import`
-1 `.accept / .input / .readFrom / .pull / .pullFrom / .reversePipe`

interface-spec.md
bergos: +1 `.match`
-1 `.filter`

interface-spec.md
pieroit: +1 `.consume / .accept / .readFrom / .reversePipe / .append`

interface-spec.md
dmitrizagidulin: +1 to `start` instead of `readable`

interface-spec.md
dlongley: I strongly recommend cribbing from:

https://nodejs.org/api/stream.html#stream_event_readable
https://streams.spec.whatwg.org/

RubenVerborgh: @dlongley We're indeed using a subset of Node.js streams (intentionally not the full interface, because its backpressure control makes it expensive).

bergos: We have at least 2 * +1 for `.import`. I will change my PR. If there are good arguments for a different method name, we can change that later.

interface-spec.md
bergos: We have at least 2 * +1 for `.match`. I will change my PR. If there are good arguments for a different method name, we can change that later.

interface-spec.md
bergos: `readable` is an [official Node.js event](https://nodejs.org/api/stream.html#stream_event_readable). So let's stay with that standard.

elf-pavlik:
IMO this PR received enough feedback to merge it after @bergos turns this feedback into new commits. After that everyone still have possibility to make new PRs proposing specific improvement to this initial draft, this should also allow for more granular changes and focused conversations.

PR #58, opened 2016-03-25
use Triple only as alias for Quads

bergos:
- removed Quad extends Triple
- Triple is just an alias of Quad
- use only the Quad interface in the spec
- removed graph parameter from .triple() factory

### DIFF
--- a/interface-spec.md
+++ b/interface-spec.md
@@ -82,10 +82,10 @@ It's only allowed to assign a `DefaultGraph` to the `.graph` property of a `Quad
 - `String .value` contains an empty string as constant value.
 
 ### Triple
-Triple is an alias of Quad.
-Triples always have `.graph` set to DefaultGraph.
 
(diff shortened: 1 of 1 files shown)


pieroit:
+1

RubenVerborgh:
Why is `graph` removed from `triple` if they are aliases?

elf-pavlik:
as i understand, .triple() **factory** returns a Quad and always assigns instance of DefaultGraph to its .graph. i added example of .triple factory which wraps .quad factory in https://github.com/rdfjs/representation-task-force/pull/51#issuecomment-198444998

pieroit:
Triple (as a class) is an alias of Quad, the factory method triple() is just going to call .quad() with DefaultGraph as fourth argument... and return a Quad

elf-pavlik:
:+1:

Issue #31, opened 2016-02-04
.literal() could have a more intelligent constructor


bergos:
We could simplify the literal constructor by detecting if the second parameter is language string or an IRI/IRI string.

`.literal('test', 'en')`
vs.
`.literal('test', 'http://example.org/datatype')`

RubenVerborgh:
:+1:

dmitrizagidulin:
:+1: to this.

PR #61, opened 2016-04-08
Clarify .literal() factory method, more intelligent constructor

dmitrizagidulin:
Addresses issue #31.

### DIFF
--- a/interface-spec.md
+++ b/interface-spec.md
@@ -15,7 +15,7 @@ This document provides a specification of a low level interface definition repre
 - Given the necessity of methods, plain objects (JSON) cannot be used.
 - Factory functions (e.g., `quad()`) or methods (e.g., `store.createQuad()`) create instances.
   - Should allow "upgrading" a plain object into a fully functional triple
-- Interfaces may have additional implementation specific properties. 
+- Interfaces may have additional implementation specific properties.
   A list of these properties maintained on the [RDFJS Representation Task Force wiki](https://github.com/rdfjs/representation-task-force/wiki/Additional-properties).
 
 ## Data interfaces
@@ -74,7 +74,7 @@ Otherwise, if no datatype is explicitly specified, the datatype IRI is `http://w
 ### DefaultGraph extends Term
 
 An instance of `DefaultGraph` represents the default graph.
-It's only allowed to assign a `DefaultGraph` to the `.graph` property of a `Quad`. 
+It's only allowed to assign a `DefaultGraph` to the `.graph` property of a `Quad`.
 
 **Properties:**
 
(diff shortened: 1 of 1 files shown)


interface-spec.md
```diff
@@ -108,11 +108,17 @@ Triple is an alias of Quad.
...
+For default values of the instance properties, see the individual [interface
+definitions](#data-interfaces)
+
 **Methods:**
 
 - `IRI .iri(String iri)` returns a new instance of IRI.
 - `BlankNode .blankNode()` returns a new instance of BlankNode.
-- `Literal .literal(String value, String language, String datatype)` returns a new instance of Literal.
+- Literal .literal(String value, [String languageOrDatatype]) returns a new
+  instance of Literal. If languageOrDatatype is an IRI, then a NamedNode is
```
nicola: how do you detect that something is an iri?

RubenVerborgh: `.termType`

nicola: so `languageOrDatatype` will have `.termType` ?

dmitrizagidulin: @RubenVerborgh I think Nicola is asking 'how do you detect what the `languageOrDatatype` parameter is, whether it's an IRI or a language code?'.

@nicola - the idea, from what I understand, is to first check if it's a language code (which I assume is easier, from a RegEx standpoint), and if it isn't, assume that it's an IRI.

bergos: The factory already allowed string values for the datatype. If it's a string the function must distinguish between languages and IRIs. The logic of the method would be something like this:

    factory.literal = function (value, languageOrDatatype) {
      if (typeof languageOrDatatype === 'string') {
        // RegExp magic
        if (isLanguage(languageOrDatatype)) {
          return new Literal(value, languageOrDatatype, langString)
        } else {
          return new Literal(value, '', this.NamedNode(languageOrDatatype))
        }
      } else {
        // superfast without detection
        return new Literal(value, '', languageOrDatatype)
      }
    }

Is it possible to implement the `isLanguage()` function in a save way? The RDF spec points to [BCP 47 2.2.9](http://tools.ietf.org/html/bcp47#section-2.2.9). Is this: `[a-z]{1,8}-[a-z]{1,8}` enough? Could this cause a conflict with an IRI?

RubenVerborgh: `/:/` seems like the fastest test for an IRI :wink:

nicola: `isLanguage` is the one I am talking about, I don't think that `/:/` is enough, what about relative iris? what about urn?

RubenVerborgh: Relative IRIs cannot occur (and we should perhaps make this clear in the spec). It is the parser's job to always return absolute IRIs. If we'd allow relative IRIs, there would be no way to distinguish them from language strings, since all language strings are valid relative IRIs.

/:/ will match on any valid absolute IRI, including URNs (which need in fact two colons). It will not match on a language string. Hence, it is an acceptable test.

nicola: does this mean that I can't have triples with relative uris?

RubenVerborgh: There's no such thing as triples with relative URIs.

Syntaxes can write triples with relative URIs, but triples themselves always have (absolute) URIs.

nicola: Oh alright, I think we are on the same page. In my syntax I can leave this relative and whatever parser will make sure to resolve the uri

elf-pavlik: Even in your syntax I don't think you will often use relative IRIs for **datatype** of the Literal term type. This issue doesn't talk about IRIs which denote `s p o g` but IRIs used for datatypes, most often `xsd:` https://www.w3.org/TR/rdf11-concepts/#section-Datatypes

blake-regalia:
Instead of requiring implementers to test the `languageOrDatatype` string parameter with regex or substring matching; the low-level API should require that datatypes are passed as a Term object, and `string` is reserved for language tags. As pointed out in the conference call, even in "bad practice", RDF formats do allow (and in practice, some users in fact do) use IRIs that don't contain colon characters.

If an implementer wants to make their API flexible at the cost of extra testing/sanitizing, and allow datatypes to passed as a string, that is their prerogative. The low-level spec should enforce strict typing for the intended use of an argument (i.e., `string` implies language-tag, `object` Term implies datatype)

niklasl:
+1 on enforcing strict typing in the low-level, as @blake-regalia writes.

PR #63, opened 2016-04-08
Rename .termType to .valueType

dmitrizagidulin:
Per @bergos - The Term can be seen as a container. `.termType` would
mean the type of the Term, so the interface name/type. But it
describes the type of the value ("iri", "bnode", "literal", and
"variable"). We had a very long discussion about these values. So
let's keep these values and rename `.termType` to `.valueType`. Than
it would describe the type of the value the container holds.

### DIFF
--- a/interface-spec.md
+++ b/interface-spec.md
@@ -15,7 +15,7 @@ This document provides a specification of a low level interface definition repre
 - Given the necessity of methods, plain objects (JSON) cannot be used.
 - Factory functions (e.g., `quad()`) or methods (e.g., `store.createQuad()`) create instances.
   - Should allow "upgrading" a plain object into a fully functional triple
-- Interfaces may have additional implementation specific properties. 
+- Interfaces may have additional implementation specific properties.
   A list of these properties maintained on the [RDFJS Representation Task Force wiki](https://github.com/rdfjs/representation-task-force/wiki/Additional-properties).
 
 ## Data interfaces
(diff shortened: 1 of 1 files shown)


RubenVerborgh:
It describes more than the type of the value in the case of literal: it also determines the presence of the `.datatype` and `.language` attributes. I find `.valueType` confusing with regard to a literal's value and datatype.

bergos:
Yes, we discussed this already in the meeting, and it's not perfect for literals, but if we keep it, the valid values are confusing. If Term is the container, `.termType` is the type of the container and the container is no longer an IRI.

RubenVerborgh:
> the container is no longer an IRI

It never was; we knew that in advance. `'iri'` is just more convenient than `'namednode'` to type.

dmitrizagidulin:
I guess it comes down to convenience (`'iri'` is shorter/easier) versus consistency / potential confusion?

l00mi:
How about getting the other way and call this property only "term" ?
That we fill it with the type of the term is kind of obvious, or isn't it ?

elf-pavlik:
Since the spec already names specific types of terms:
* NamedNode
* BlankNode
* Literal
* Variable
* DefaultGraph

Also used as types of what [DataFactory methods](https://github.com/rdfjs/representation-task-force/blob/master/interface-spec.md#datafactory) return, why not just stick to those 5 strings and this way avoid introducing additional names for term types, with just slightly different spelling.

* iri
* bnode
* literal
* variable
* defaultGraph

bergos:
I proposed that already last time we discussed this, but we discussed this already. We should avoid endless loops :-(

If people who voted for the short string values change there mind, I'm happy to change the spec, but also I would like to close this topic.

Arguments pro value == interface/factory name:

- avoid introducing new names
- RDF-Interfaces already also uses values like that

elf-pavlik:
https://www.w3.org/TR/rdf-interfaces/#basic-node-types
> interfaceName of type DOMString, readonly

>    Provides access to the string name of the current interface, normally one of "NamedNode", "BlankNode" or "Literal".

>    This method serves to disambiguate instances of RDFNode which are otherwise identical, such as NamedNode and BlankNode.

After merging various open PRs I'll make a new one to propose following rdf-interfaces spec and forward it to mailing list. 


:-1: to changing .termType to .valueType

elf-pavlik:
my alternative proposal https://github.com/rdfjs/representation-task-force/pull/65

bergos:
Closed, because we merged [#65](https://github.com/rdfjs/representation-task-force/pull/65).

PR #65, opened 2016-04-28
use interface name for .termType value

elf-pavlik:
Using interface name for value of .termType follows approach used by https://www.w3.org/TR/rdf-interfaces/#basic-node-types and implemented in rdf-ext

By reusing interface names we avoid introducing additional special symbols for identifying the type of interface.

This PR also proposes alternative to currently open #63

### DIFF
--- a/interface-spec.md
+++ b/interface-spec.md
@@ -28,7 +28,7 @@ Abstract interface.
 
 **Properties:**
 - `String .termType` contains a value that identifies the concrete interface of the term, since Term itself is not directly instantiated.
-  Possible values include `"iri"`, `"bnode"`, `"literal"`, and `"variable"`.
+  Possible values include `"NamedNode"`, `"BlankNode"`, `"Literal"`, `"Variable"` and `"DefaultGraph"`.
 - `String .value` is refined by each interface which extends Term
 
 **Methods:**
@@ -43,21 +43,21 @@ Abstract interface.
 
 **Properties:**
 
-- `String .termType` contains the constant `"iri"`.
+- `String .termType` contains the constant `"NamedNode"`.
 - `String .value` the IRI as a string (example: `http://example.org/resource`)
 
 ### BlankNode extends Term
 
 **Properties:**
 
-- `String .termType` contains the constant `"bnode"`.
+- `String .termType` contains the constant `"BlankNode"`.
 - `String .value` blank node name as a string, without any serialization specific prefixes, e.g. when parsing, if the data was sourced from Turtle, remove _:, if it was sourced from RDF/XML, do not change the blank node name (example: `blank3`)
 
 ### Literal extends Term
 
 **Properties:**
 
-- `String .termType` contains the constant `"literal"`.
+- `String .termType` contains the constant `"Literal"`.
 - `String .value` the text value, unescaped, without language or type (example: `Brad Pitt`)
 - `String .language` the language as lowercase [BCP47](http://tools.ietf.org/html/bcp47) string (examples: `en`, `en-gb`) or an empty string if the literal has no language.
 - `IRI .datatype` the datatype of the literal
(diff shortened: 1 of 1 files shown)


dmitrizagidulin:
+1 from me, seems like a better approach.

bergos:
I expect we don't have to type `node.termType === 'something'` many times because high level libraries will add methods like `.isBlankNode`, `.is...`

RubenVerborgh:
I've always seen the strings as a shortcut. If they're not a shortcut, the point of having them perhaps changes. Should we switch to `instanceof`, or is this too limited?

bergos:
We can't use `instanceof`, because it's up to the factory which implementation it returns.

RubenVerborgh:
…but it forces the implementation to use actual prototypes.

bergos:
It's just a string value with the name of the interface, not the prototype.

elf-pavlik:
> Instances of the interfaces created with different libraries should be interoperable. 

As I understand, a runtime may have multiple object, each of which implements the interface of certain term type (eg. Literal) but instantiated using different prototypes. For that reason we can't use `instanceof` to check the type of implemented interface.

If I understand it correctly, I would like to create separate issue to give it more emphasis in the spec and possibly mention reason for not using `instanceof` explicitly.  

Also to draw stronger distinction between name/type of the interface and name/type of one of many prototypes which implement that interface, we could follow rdf-interfaces spec and use property `interfaceName` or `interfaceType`?

Data factory section uses "returns a new instance of {termType}" all the time. Maybe we should change it to "returns a new instance with {termType} interface".

elf-pavlik:
based on https://github.com/rdfjs/representation-task-force/pull/63#issuecomment-210113353

| Interface Name | termType | factory |
----------------------|------------------- |------
NamedNode | iri | namedNode
BlankNode | bnode | blankNode
Literal | literal |literal
Variable | variable | variable
DefaultGraph | defaultGraph | defaultGraph

IMO the special case of *iri* and *bnode* seems unnecessary and forces people to remember two extra magic strings. If not interface name, we could use string matching factory name, as currently for *literal*, *variable* and *defaultGraph*.

pieroit:
I was for `IRI` much more than `NamedNode`, but the consistency between Interface Name, termType and factory is more important.

Either we go for `IRI | 'iri' | iri()` **or** `NamedNode | 'namedNode' | namedNode()`.
Mixing the two is just confusing.

elf-pavlik:
If people prefer lowercase string (same as factory method names) over uppercase strings (which we use for naming subclasses of Term) I can add commit to adjust that.
I mostly want to remove the special case of `iri` and `bnode`. Since we check .term**Type** upper case version seemed more intuitive to me.

elf-pavlik:
https://github.com/rdfjs/representation-task-force/blob/master/interface-spec.md
>
- `NamedNode .namedNode(String iri)` returns a new instance of NamedNode.
- `BlankNode .blankNode([String identifier])` returns a new instance of BlankNode.
  The optional identifier parameter is assigned to `.value`.
  If the label parameter is undefined a new value is generated for each call.
- `Literal .literal(String value, [String languageOrDatatype])` returns a new
  instance of Literal. If languageOrDatatype is an IRI, then a NamedNode is
  constructed with that IRI, and is used for the value of `.datatype`.
  Otherwise languageOrDatatype is used for the value of `.language`.
- `Variable .variable(String name)` returns a new instance of Variable. This method is optional.
- `DefaultGraph .defaultGraph()` returns an instance of DefaultGraph.

Currently in this PR, value of `.termType` matches exactly the string identifying which type of Term each factory method returns. This seems very intuitive to me and once again follows how rdf-interfaces spec defined it.

elf-pavlik:
@l00mi pointed out during the call that if we need to change it based on implementation experience to shorter strings, people implementing it can do simple search & replace.

PR #66, opened 2016-04-28
rename factory parameters to value

elf-pavlik:
Since values of parameters of factory methods get assigned to `.value`. Naming them *value* seems more intuitive where person will expect to access passed value as `.value`. We also avoid introducing additional terminology into the spec.

### DIFF
--- a/interface-spec.md
+++ b/interface-spec.md
@@ -109,8 +109,8 @@ Triple is an alias of Quad.
 
 ### DataFactory
 
-For default values of the instance properties, see the individual [interface
-definitions](#data-interfaces)
+For default values of the instance properties and valid values requirements,
+see the individual [interface definitions](#data-interfaces)
 
 **Methods:**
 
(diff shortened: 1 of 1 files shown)


elf-pavlik:
This way we also can keep requirements for the value in one place. The `.value` definition specific to each type of term. I can change sentence
> For default values of the instance properties, see the individual interface definitions

into something in lines:

*For default values of the instance properties and requirements of valid values, see the individual interface definitions*

elf-pavlik:
Blank Node seems like a good example where delegating to interface definition makes a lot of sense
> `String .value` blank node name as a string, without any serialization specific prefixes, e.g. when parsing, if the data was sourced from Turtle, remove `_:`, if it was sourced from RDF/XML, do not change the blank node name (example: `blank3`)

Based on just *identifier* one could try to use `_:`

dmitrizagidulin:
I'm happy to change my -1 to a non-blocking -0.5, in the interest of consensus.

(I still think that it makes the spec harder for developers to use at a glance. But I can live with it).

Issue #79, opened 2016-07-07
Forward DataFactory to parsers/readers


bergos:
The constructor of readers should have an option to define the DataFactory that is used to create Terms and Quads.

Proposal: Use an object as last parameter of the constructor with the property  `factory`

elf-pavlik:
In case of RDFa one may also need to pass a `@base` to resolve relative IRIs
ping @alexmilowski @csarven

EDIT: moved to #96

RubenVerborgh:
In any case, make `baseIRI` and `factory` properties of an options object, not individual parameters.

bergos:
The libraries I contributed are using the `factory` option. I saw [rdfxml-streaming-parser.js](https://github.com/rdfjs/rdfxml-streaming-parser.js) is using `dataFactory` and I think also some others are using `dataFactory`. Something we should discuss again. But would be easy to support both till the next major version.

k00ni:
I am arguing from a PHP developer perspective. I'd rather use different parameter instead of an options object. IMHO this pattern is only valid in some cases, it may cloud its structure, if the number of options rise and makes type hints harder to use.

But i can adapt to the specification, if it requires an options object.

RubenVerborgh:
Options are idiomatic in JavaScript. It might not be a good idea to follow this spec for another language, where other things are idiomatic.

elf-pavlik:
I believe https://fetch.spec.whatwg.org/#fetch-method makes a good example with its second optional parameter https://fetch.spec.whatwg.org/#requestinfo

Issue #44, opened 2016-02-25
Optional options parameter for all source, sink and store methods


bergos:
Many source, sink and store implementations may require addition options like base, HTTP headers (e.g. ETag), credentials and many more. We should add the optional `Object options` parameter in the spec. The `base` property should be also described in the spec.

RubenVerborgh:
This could just be a property of their constructor, which we don't need to
specify I think.

bergos:
The constructor can and should be used for most options, but for some properties, it doesn't make sense.

The `base` option could be set in the constructor, because usually creating a new instance of a parser is very lightweight. But I would expect the option in the parse (`.read`) method.

I implemented a store for access control, that just wraps another store. Creating the access control store isn't lightweight. I'm doing some caching and the cache is attached to the store. The user (WebID) is forwarded to the store using the options object. If only the constructor can be used to set the options I need to know the implementation and additional properties. Usually I do that in the configuration. A fork method would be an option, but also doesn't look very nice:

with options:

    store.read(null, null, null, {user: req.session.user}).pipe(res)

with constructor:

    var store = new conf.Store(_.defaults(conf.storeOptions), {user: req.session.user})

    store.read().pipe(res)

with fork:

    var s = 'fork' in store ? store.fork({user: req.session.user})

    s.read().pipe(res)

It easy to add a new implementation specific parameter, but I would like to reserve the parameter to make sure it's used in the same way in all implementations. Beside the base I would handle it like the additional properties for Quads.

dmitrizagidulin:
+0 to adding `options` parameter, no strong opinion.

RubenVerborgh:
-1, I heavily object. It's bad design if a user of `store` needs to pass an optional options object, needs to know whether to pass that object, and if so, what it should contain. It's a haystack being passed around (see [Law of Demeter](http://googletesting.blogspot.be/2008/07/breaking-law-of-demeter-is-like-looking.html)).

This is the perfect option:
var store = new conf.Store({user: req.session.user});

store.read().pipe(res);
```

(minus the defaults; that the store should do itself) because construction can be separated from usage.

> Creating the access control store isn't lightweight.

[Creating objects should always be lightweight.](http://misko.hevery.com/code-reviewers-guide/flaw-constructor-does-real-work/). They should only be initialized on actual use; e.g., when a call to `read` is made. That does not mean the parameter should be passed to read.

bergos:
[...]

PR #80, opened 2016-07-19
added Reader and Writer interfaces

bergos:


RubenVerborgh:
Note sure about the necessity of writer. Instead of writing to something, that thing could just read from you.

bergos:
That would require to have a generic stream as return value and parameter. Should we define `Stream` without specifying the data type? `QuadStream` could inherit from `Stream` with a specific data type and be used everywhere in the spec where we have written `Stream` until now.

RubenVerborgh:
Since JavaScript is ducktyped, it doesn't really matter; how we call it is just for the documentation. Perhaps simply `Stream<Quad>` to refer to it.

elf-pavlik:
Any updates needed on this one?

elf-pavlik:
green-turtle uses term *Processor*, since we may want to avoid confusion with node *Transform* and as I understand implementations will use it for both parsers and serializers, *Processor* may describe this *stream translator* better than *Reader*

elf-pavlik:
This PR seems to miss definition of a constructor/factory, #79 suggests that constructor would allow passing factory, possibly one could also pass base as in #96

elf-pavlik:
@bergos what do you think about changing /Reader/Processor/ & /read()/process()/ ?
next PR could address options to pass `base`, `@context` for JSON-LD serializer etc.

### Processor

A Processor is an object that translates one stream to another stream.
The input and output stream may emit quads or custom formats.
The Processor interface can be used for parsers and serializer, but also for quad or custom format transformations.
Implementations may support addition data types like String or Buffer as input

`Stream .process(Stream input)`
Creates a `Stream` which emits data based on the `input` stream.

timbl:
If, when parsing a turtle document, or within an N3 document, you are parsing a sub graph bewteen { braves } you create a stream of triples  then do you also create a stream of nodes when you are a parsing a Collection/List/Array between ( parentheses) ?

RubenVerborgh:
@timbl I'd say that everything is a stream of triples (for consistency). So
parsing a list leads to a stream of rdf:first/rdf:rest. That's how current
streaming parsers in JavaScript work. Would this be problematic for N3,
which gives lists a first-class status? In any case, switching types
mid-stream seems difficult.

(Different but perhaps related: I recently was in touch with @josd
regarding the equal treatment of lists In EYE from N3/Turtle/N-Triples
input, so no special treatment seems necessary.)

elf-pavlik:
> @timbl: If, when parsing a turtle document, or within an N3 document, you are parsing a sub graph bewteen { braves } you create a stream of triples

> @RubenVerborgh: I'd say that everything is a stream of triples (for consistency).

Parser should return stream of Quads, (where we have Triple as alias of quad with .graph always set to DefaultGraph). AFAIK @blake-regalia trig/n-quads parser in graphy.js streams quads as well as @RubenVerborgh's trig/n3/n-quads parser in N3.js

I must admit not having much familiarity with N3, I don't know how 'sub graph' in N3 differs from 'named graph' in Trig/N-Quads. I also remember @RubenVerborgh [asking very specific question about N3 lists in a chatroom](https://gitter.im/rdfjs/Representation-Task-Force?at=580e21ac9aea834129e9fbc2) which possibly @timbl never answered there. If N3 has some additional requirements besides requirements of all the [official RDF1.1 serializations](https://www.w3.org/TR/rdf11-new/#section-serializations) maybe it would make sense to start documenting them in dedicated .md file or on wiki page? This would also help everyone to better understand N3 :smile:

bergos:
After playing a little bit with the interfaces, I don't think we need to distinguish between reader, writer or processor. We have already the `.import` method of the `Sink` interface. With [that PR](https://github.com/rdfjs/representation-task-force/pull/110) it will return an EventEmitter. I think it's required to track the status of the import. That PR has already some +1s.

`EventEmitter .import(Stream stream)` should cover everything we need. `EventEmitter` can be also a `Stream`. `Stream` is not exclusive for `Quad`, but can be also used for byte and object streams. Here is list of use cases and subtypes, which are all covered by the method:

**Parser**
`Stream<Quad>|Source .import(Stream stream)`

**Serializer**
`Stream .import(Stream<Quad>|Source stream)`

**Transformation**
`Stream<Quad>|Source .import(Stream<Quad>|Source stream)`

**Store**
`EventEmitter .import(`EventEmitter .import(Stream stream)` stream)`

Instead of adding more interfaces, I would propose to change the description of the method. Optional we can even mention the subtypes explicitly.

If we should close this PR and create a new one with the extended description, please +1 on this comment.

elf-pavlik:
https://github.com/rdfjs/representation-task-force/blob/master/interface-spec.md#sink
> A Sink is an object that consumes quads. It can store the quads or do some further processing. For example serializers and transformations which consume quads can implement the Sink interface.
>*  undefined .import(Stream stream) Writes all quads from the stream to the sink.

So `.import` would accept any Stream, and for Store we should clarify that in only accepts `Stream<Quad>` (just like serializer and transformation)

elf-pavlik:
Noticing #79 and how rdf-ext implements it https://github.com/rdf-ext/rdf-ext/blob/develop/lib/parsers.js#L33

It seems that `import()` should also accept `options` parameter, which also has long discussion in #44

Issue #89, opened 2016-09-29
Simplify canonical form for performance


RubenVerborgh:
The canonical form needs to be fast, because it is the unique key for each term. As such, it is the prime candidate for indexing etc., which is a very common task (triple stores, query engines).

The current canonical form requires escaping, which is horribly slow and brings the computation cost of `toCanonical` at log(n) with n the length of the value. (This can be cached, but then memory size doubles.)

I propose a much more simple canonical form, for instance what I [am using here](https://github.com/RubenVerborgh/N3.js#triple-representation).

RubenVerborgh:
With regard to aesthetic arguments, I don't mind if `toString` (which is what developers ultimately see) does something fancy or expensive. Just not in a performance-critical function like canonical form.

bergos:
Indexing can be also done with an integer and a lookup table. It depends on the use case and should be therefore implementation specific. This can be done with an addition property or method.

The current `toCanonical` method is based on the N-Triples spec. Other specs depend already on it. We should not reinvent something new.

l00mi:
I like the idea to optimize for speed based on javascript built-in mechanisms (in contrast to optimize with new datastructures) for such an important function. Why not adding a additionally an toN3 which goes by the specs. Like this the API says implicit that the toCanonical has to be something else.

RubenVerborgh:
> Indexing can be also done with an integer and a lookup table.

How are you going to get that integer? Based on the canonical representation… which is too slow.

> We should not reinvent something new.

We're already doing that anyway, because N-Triples does not provide variables, quads, default graph.

RubenVerborgh:
> Why not adding a additionally an toN3 which goes by the specs. Like this the API says implicit that the toCanonical has to be something else.

It seems easier to just get `toCanonical` right :wink:

elf-pavlik:
> The current canonical form requires escaping, which is horribly slow and brings the computation cost of toCanonical at log(n) with n the length of the value.

@RubenVerborgh could you possibly explain a bit more about this slow escaping? Does it only affect Literal or also other types of Term?

RubenVerborgh:
> could you possibly explain a bit more about this slow escaping?

Assuming that the in-memory representation contains only an unescaped version, you need to search and replace every time a canonical version is readed.

> Does it only affect Literal or also other types of Term?

Only Literal.

elf-pavlik:
[...]

Issue #88, opened 2016-09-29
Define equals using canonical representation


RubenVerborgh:
At the moment, equality is defined as:

> `boolean .equals(Term other)` returns `true` if and only if the argument is a) of the same type b) has the same contents (value and, if applicable, type or language)

I propose to rewrite this as

> `boolean .equals(Term other)` returns `true` if and only if the argument has the same canonical representation

for the following reasons:
- “of the same type” is not correct (it implies `instanceof` comparisons), and should be “has the same term type”
- both (the corrected) a) and b) are consequences of the canonical representation
- there is no way to exhaustively test for “same contents”

elf-pavlik:
Should spec include some note about what equality means for blank nodes? Thinking about Gregg's comment in https://github.com/rdfjs/representation-task-force/issues/50#issuecomment-201698108

> for BlankNode being the manifest label doesn't hold up, as if two graphs are merged, blank nodes with the same label aren't merged, as they are different nodes, even if they are labeled the same in a concrete serialization used to instantiate the graph. A graph holds an abstract representation created by parsing a concrete representation into terms; blank node labels are just convenient for a concrete serialization, even if some vestige remains as a graph term.

I can't recall if two different blank nodes in the same dataset can share the same identifier/label

RubenVerborgh:
> Should spec include some note about what equality means for blank nodes?

Yes. It's up to the parser to generate new blank node labels when parsing two different documents that use the same label.

> I can't recall if two different blank nodes in the same dataset can share the same identifier/label

They cannot share in any of the standard RDF syntaxes to serialize a dataset.

bergos:
We had this already [in this PR](https://github.com/rdfjs/representation-task-force/pull/53). We can change it to use `termType` instead of type, so nobody implements it with `instanceof`. But we should keep the rest like it is.

A factory should create unique blank node labels, if the identifier is not given. This should be the default behavior. Reusing the identifier should be an option that must be explicitly enabled.

RubenVerborgh:
> We can change it to use termType instead of type, so nobody implements it with instanceof.

+1

> We had this already in this PR.

Note that my argument was there: `given that .toCanonical might need to perform a computation`; however, if #89 is resolved, that argument might become invalid (all the more if `toCanonical` becomes a property).

So, I think this issue then depends on the solution for #89.

RubenVerborgh:
[...]

PR #90, opened 2016-10-01
Remove toCanonical()

RubenVerborgh:
I propose to remove `toCanonical`, since there is currently no inherent need to have it. It is not necessary for equality comparisons at the moment (#88, #53). The current definition is too expensive for repeated use, for instance for keying (#89).

Furthermore, `toCanonical` is not a "core" property of a triple/quad/term like _subject_, _predicate_, _object_, _value_, _datatype_, _language_, _termType_ are. So we should really wonder whether it makes sense to require all implementers to have it.

Libraries can implement their own `toCanonical`, `toString`, or any other display method; there is no need to agree on it (and as such, it shouldn't be part of the spec).

So unless we really can identify an added value for having a cross-implementation `toCanonical`, the more logical thing might be to simply remove it.

### DIFF
--- a/interface-spec.md
+++ b/interface-spec.md
@@ -95,8 +95,6 @@ Triple is an alias of Quad.
 - `Term .object` the object, which is an IRI, a Literal, a BlankNode or Variable.
 - `Term .graph` the named graph, which is an IRI, DefaultGraph, BlankNode or Variable.
 
-- `Boolean .equals(Quad other)` returns true if and only if the argument is a) of the same type b) has all components equal
-
 **Methods:**
 
 - `string .toCanonical()` returns a canonical string representation of the quad.
--- a/interface-spec.md
+++ b/interface-spec.md
@@ -34,10 +34,6 @@ Abstract interface.
 **Methods:**
 
 - `boolean .equals(Term other)` returns true if and only if the argument is a) of the same type b) has the same contents (value and, if applicable, type or language)
-- `string .toCanonical()` returns a canonical string representation of the term.
-  For IRIs, BlankNodes and Literals the [N-Triples canonical form](https://www.w3.org/TR/n-triples/#canonical-ntriples) must be used.
-  Variables must return the variable name prefixed with a question mark (example: `?a`).
-  DefaultGraph must return an empty string.
 
 ### NamedNode extends Term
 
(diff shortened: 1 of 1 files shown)


RubenVerborgh:
It might be part of the high-level library (but there, methods such as `toNT` might make even more sense).

interface-spec.md
```diff
@@ -95,17 +91,9 @@ Triple is an alias of Quad.
 - `Term .object` the object, which is an IRI, a Literal, a BlankNode or Variable.
 - `Term .graph` the named graph, which is an IRI, DefaultGraph, BlankNode or Variable.
 
-- `Boolean .equals(Quad other)` returns true if and only if the argument is a) of the same type b) has all components equal
-
 **Methods:**
```
elf-pavlik: PR title only mentions removal of toCanonical() and here you remove equals() on Quad

RubenVerborgh: This was an incorrect duplicate (note how it is listed under properties), already fixed in master. This PR should be rebased.

RubenVerborgh: Turns out I hadn't pushed master yet. Done now.

blake-regalia:
I wholeheartedly support this. No need for canonicalization to be in an interface spec. If anything, _serialization_ could be part of a higher-level spec.

elf-pavlik:
👍 from https://gitter.im/rdfjs/Representation-Task-Force?at=57f142156f39416b1ce11057

timbl:
Two notes 
1) when worrying about the speed of escaping,  then sure it is linear -- but linear in what?  Linear in the length of a data string. In my systems I note most strings are small - the graph has lots of small strings in it, and a few big one 
2) One could specify the toCanonical function but NOT specify the algo --  so diff stored  could use diff forms - they just all would be consistent across the datafactory.

RubenVerborgh:
1) Linear in the number of characters; they're not tremendously big indeed, but the looping over each string is going to cost for a function that was meant to be a fast key.

2) Agree, but then perhaps it belongs in the high-level library instead. The low-level library really goes for absolute compatibility.

As a note, the way I solve it in N3 is by defining my own field `.id`, which has a fast canonical form that is only meant for internal usage and low-level keying.

Issue #83, opened 2016-08-17
 low-level API should require that datatypes as Term object and string for language


bergos:
This is a comment from @blake-regalia on this [PR](https://github.com/rdfjs/representation-task-force/pull/61#issuecomment-231149938).

> Instead of requiring implementers to test the `languageOrDatatype` string parameter with regex or substring matching; the low-level API should require that datatypes are passed as a Term object, and `string` is reserved for language tags. As pointed out in the conference call, even in "bad practice", RDF formats do allow (and in practice, some users in fact do) use IRIs that don't contain colon characters.

> If an implementer wants to make their API flexible at the cost of extra testing/sanitizing, and allow datatypes to passed as a string, that is their prerogative. The low-level spec should enforce strict typing for the intended use of an argument (i.e., `string` implies language-tag, `object` Term implies datatype)

RubenVerborgh:
:+1:

dmitrizagidulin:
Completely agree with this.

elf-pavlik:
Looks like merging #92 should resolve this issue as well

elf-pavlik:
resolved in #92

Issue #78, opened 2016-07-01
`.datatype` should be a string


blake-regalia:
### `Literal#datatype` should be a `string`

In RDF, datatypes are URIs. They are never blank nodes. Using a `Term` is a redundant and inconvenient way to make the value of the IRI a little farther from reach. As opposed to a `string`, the `Term` also has a slight negative impact on memory and performance.

#### Consistency
You may think it's consistent to wrap `.datatype` as a `Term`, but at this tier in the class relation tree, it's actually *inconsistent* with the `.value` property of a NamedNode and all other Term properties which are all `string`s.

#### `.equals`
Aside from the ontology portion of a graph, datatype IRIs just about never coincide with the IRIs of subject or object nodes (i.e. NamedNodes).
 - As a developer, it makes sense that since the overwhelming majority of datatype comparisons are to other datatypes, then the naive comparison is keenest:  `if(lit1.datatype === lit2.datatype)`, rather than the obscure cases where having datatype already be a `Term` comes in handy:

    ```js
    let subject = // where datatype also happens to appear as a subject
    if(lit1.datatype.equals(subject))
    ```

   - Rather, for those rare incidents, you can do:
 
    ```js
    if(DataFactory.namedNode(lit1.datatype).equals(subject))
    // -- or, better yet --
    if(subject.value === lit1.datatype && isNamedNode(subject))
    ```

    - Again, these are obscure cases - usually to fetch some information about the datatype itself from the ontology.
 - The much more common application of datatypes is for testing whether or not a particular datatype is the right one.

    ```js
    // right now, the implication is to use the interface method:
    if(lit1.datatype.equals(DataFactory.namedNode('http://www.opengis.net/ont/geosparql#')))
    // -- but that is messy, less elegant, and much slower perfomance-wise than --
    if(lit1.datatype === 'http://www.opengis.net/ont/geosparql#')
    ```

bergos:
#### Consistency

The other properties are `string`s, because of the mapping of the RDF Spec to JS types. Using `NamedNode` whenever there is a `NamedNode` would be more consistent.

#### .equals

If we add CURIE support in the future, calling `.equals` would be much more elegant. Also, this is a low level API. High level libraries may implement a simpler API.

There is also a simpler way doing a string compare you don't show in your examples:

    if(lit1.datatype.value === 'http://www.opengis.net/ont/geosparql#')

bergos:
[...]

Issue #50, opened 2016-03-02
Proposal: Rename IRI node type to NamedNode


dmitrizagidulin:
Per yesterday's conversation on rdfjs public gitter channel -- Propose to rename the `IRI` node type on the [Data Interfaces spec](https://github.com/rdfjs/representation-task-force/blob/master/interface-spec.md) to `NamedNode` (to better match `BlankNode`).

RubenVerborgh:
Thanks for creating this issue. There has been a long discussion about this in previous meetings, and we settled on IRI because this is the term actually used by developers. This is a library for developers, we should speak their language. (Summary of a very long discussion.)

Given that this discussion has been had previously, I'm hesitant to reopen this, unless there are new arguments. Could you link to / summarize the discussion on the gitter channel?

RubenVerborgh:
See also #1 and reports of the first few meetings.

timbl:
"this is the term actually used by developers"  -- but for something different!

timbl:
Yes, I am sorry that the long email discussion on the list was decided to support "IRI".  I don't think the argument was that developers us it.   I thought that the prevailing reason was that the term "IRI" was used by the RDF 1.1 spec, and that spec carried a lot of weight.  Which is unfortunate, as I think it will confuse developers and was a mistake in the rdf1.1 spec.

RubenVerborgh:
As far as I recall, the deciding argument was to make it easy for developers; the spec was one of the other reasons. If indeed the specifications themselves conflate (purposely, for simplicity) IRIs and named nodes, it's very hard to explain this to developers.

However, I agree the correct term is "named node" and that an IRI is just a string. Yet we should be able to point developers to a few official places where "named node" and IRI are used correctly; it would be hard to take the task upon ourselves to make that definition. So can we borrow the correct terminology from somewhere?

Also, we should remember that `iri` itself would only be used in very few places in any RDF/JS library. The interface name itself is never exposed; the only instances are `.termType` (which currently returns `"iri"`) and the constructor name. We could define an alias for the constructor name, so both `iri` and `namedNode` are support; we can, however, only define one return value for `.termType` (and `"iri"` is handy because it just has 3 letters).

elf-pavlik:
I guess people developing for the web in JS may have some familiarity with https://nodejs.org/api/url.html

[...]

PR #92, opened 2016-10-02
Replace obsolete mentions of "IRI"

RubenVerborgh:
The specification still mentioned "IRI" in a couple of places, where actually "NamedNode" was intended. This pull request fixes this.

For clarity, here are the definitions:
- an IRI is a string
- `NamedNode` is the RDF/JS interface to represent nodes identified by an IRI

There is one case for which I'm in doubt, which should be **resolved in #93 _before_ this pull request can be accepted**.

### DIFF
--- a/interface-spec.md
+++ b/interface-spec.md
@@ -8,7 +8,7 @@ This document provides a specification of a low level interface definition repre
 **Purpose:** This document proposes a uniform interface to represent RDF data in low-level JavaScript libraries.
 
 ## Design elements and principles
-- We define data interfaces to represent **triples**, **quads**, **IRIs**, **blank nodes**, **literals** and **variables**.
+- We define data interfaces to represent **triples**, **quads**, **named nodes**, **blank nodes**, **literals** and **variables**.
 - Instances of the interfaces created with different libraries should be interoperable.
 - Interfaces do _not_ specify how instances are stored in memory.
 - Interfaces mandate specific pre-defined methods such as `.equals()`.
@@ -40,7 +40,7 @@ Abstract interface.
 **Properties:**
 
 - `string .termType` contains the constant `"NamedNode"`.
-- `string .value` the IRI as a string (example: `http://example.org/resource`)
+- `string .value` the IRI of the named node (example: `http://example.org/resource`)
 
 ### BlankNode extends Term
 
@@ -56,10 +56,10 @@ Abstract interface.
 - `string .termType` contains the constant `"Literal"`.
 - `string .value` the text value, unescaped, without language or type (example: `Brad Pitt`)
 - `string .language` the language as lowercase [BCP47](http://tools.ietf.org/html/bcp47) string (examples: `en`, `en-gb`) or an empty string if the literal has no language.
-- `IRI .datatype` the datatype of the literal
+- `NamedNode .datatype` a NamedNode whose IRI represents the datatype of the literal
 
-If the literal has a language, the datatype IRI is `http://www.w3.org/1999/02/22-rdf-syntax-ns#langString`.
-Otherwise, if no datatype is explicitly specified, the datatype IRI is `http://www.w3.org/2001/XMLSchema#string`.
+If the literal has a language, its datatype has the IRI `http://www.w3.org/1999/02/22-rdf-syntax-ns#langString`.
+Otherwise, if no datatype is explicitly specified, the datatype has the IRI `http://www.w3.org/2001/XMLSchema#string`.
 
 ### Variable extends Term
 
@@ -86,10 +86,10 @@ Triple is an alias of Quad.
 
 **Properties:**
 
-- `Term .subject` the subject, which is an IRI, a BlankNode or Variable.
-- `Term .predicate` the predicate, which is an IRI or Variable.
-- `Term .object` the object, which is an IRI, a Literal, a BlankNode or Variable.
-- `Term .graph` the named graph, which is an IRI, DefaultGraph, BlankNode or Variable.
+- `Term .subject` the subject, which is a NamedNode, BlankNode or Variable.
+- `Term .predicate` the predicate, which is a NamedNode or Variable.
+- `Term .object` the object, which is a NamedNode, Literal, BlankNode or Variable.
+- `Term .graph` the named graph, which is a DefaultGraph, NamedNode, BlankNode or Variable.
 
 **Methods:**
 
@@ -106,8 +106,7 @@ see the individual [interface definitions](#data-interfaces)
 - `BlankNode .blankNode([string value])` returns a new instance of BlankNode.
   If the value parameter is undefined a new identifier for the blank node is generated for each call.
 - `Literal .literal(string value, [string languageOrDatatype])` returns a new
-  instance of Literal. If languageOrDatatype is an IRI, then a NamedNode is
-  constructed with that IRI, and is used for the value of `.datatype`.
+  instance of Literal. If languageOrDatatype is a NamedNode, then it is used for the value of `.datatype`.
   Otherwise languageOrDatatype is used for the value of `.language`.
 - `Variable .variable(string value)` returns a new instance of Variable. This method is optional.
 - `DefaultGraph .defaultGraph()` returns an instance of DefaultGraph.
@@ -182,6 +181,6 @@ Access to stores LDP or SPARQL endpoints can be implemented with a Store intefac
   All quads matching the pattern will be removed.
   The `end` and `error` events are used like described in the `Stream` interface.
 
-- `EventEmitter .deleteGraph(IRI|string graph)`
+- `EventEmitter .deleteGraph(Term|string graph)`
   Deletes the given named graph.
   The `end` and `error` events are used like described in the `Stream` interface.


interface-spec.md
```diff
@@ -60,10 +60,10 @@ Abstract interface.
 - `string .termType` contains the constant `"Literal"`.
 - `string .value` the text value, unescaped, without language or type (example: `Brad Pitt`)
 - `string .language` the language as lowercase [BCP47](http://tools.ietf.org/html/bcp47) string (examples: `en`, `en-gb`) or an empty string if the literal has no language.
-- `IRI .datatype` the datatype of the literal
+- `string .datatype` the IRI of the literal
```
elf-pavlik: how about 
`string .datatype` the IRI denoting the datatype of literal

RubenVerborgh: Certainly, yes! I went to quick there.

interface-spec.md
```diff
@@ -192,6 +191,6 @@ Access to stores LDP or SPARQL endpoints can be implemented with a Store intefac
   All quads matching the pattern will be removed.
   The `end` and `error` events are used like described in the `Stream` interface.
 
-- `EventEmitter .deleteGraph(IRI|string graph)`
+- `EventEmitter .deleteGraph(Term|string graph)`
```
elf-pavlik: can this one use more specific NamedNode instead of Term? or we also want to use it to delete the DefaultGraph?

RubenVerborgh: Graphs can also be blank nodes.

elf-pavlik: on second thought Generalized RDF allows blank nodes as graph names...

interface-spec.md
RubenVerborgh: Fixed in cfcfa967965f813b5051128b74de9505f6b46ce1.

RubenVerborgh:
This PR has been updated to incorporate the resolution from #93 / #78.

retog:
In the RDF abstract syntax the elements of a triple may be "IRIs, blank nodes or datatyped literals", there is no mention of "named nodes". To me it would make more sense to use the terminology from RDF wherever possible. So this would mean adapting in the other direction.

RubenVerborgh:
@retog We had extremely long discussions on this, with @timbl and others. This is considered sloppiness of the specs. An IRI is a string; a named node is not, hence our choice for named node. See #50 for details.

elf-pavlik:
I think without distinct definitions of NamedNode (Term object) and IRI (string primitive) we would have really hard time to clarify issues #78 & #83 

@bergos this one seems ready to merge and after that I think we can already close both issues I mention in the line above.

retog:
@RubenVerborgh, I don't see the sloppiness in the spec, on the contrary the double indirection IRI -> Node -> Resource doesn't make sense, imho. But I see it was discussed and the decision was made despite @gkellogg already pointing out the inconsistency of this approach, the API will hopefully be usable anyway.

RubenVerborgh:
@retog  The sloppiness in the spec is that an IRI is a string, but the spec conflates IRI and node.

> the double indirection IRI -> Node -> Resource

There's no double indirection: it is IRI -> Named Node (and Named Node is a subclass of Term). Does that make it better? (Do you have a pointer to @gkellogg's remark?)

retog:
Here's the comment: https://github.com/rdfjs/representation-task-force/issues/50#issuecomment-201698108

The indirection is double because the Named Node then somehow points to an element in the universe. According to RDF semantics it is the interpretation function that delivers the element in the universe given a specified IRI or literal. This is not so for blank nodes, only names (i.e. IRI or literal) have a referent, blank nodes only say that a thing exists. With the named nodes there is a  to 1 relation from IRIs to NamedNodes (I assume this is how you intend it to be) and a many to one relation from NamedNodes to Resources (i.e. things in the universe). As I said, I consider the indirection with the NamedNodes a completely unnecessary addition of complexity and detachment from the specs. As @gkellogg pointed out it is also inconsistent that no such intermediate entity id created for the Literals, by RDF 1.1 Semantics both IRIs and Literals are names.

RubenVerborgh:
> With the named nodes there is a to 1 relation from IRIs to NamedNodes

Honestly, I don't know how to counter that argument, as it was my own argument as well at the very beginning of the discussion (I also used to say that we don't need named nodes). See [here](https://github.com/rdfjs/representation-task-force/wiki/NamedNode#conjecture-1-there-exists-a-one-to-many-relationship-from-named-node-to-iri) for the relevant part of my long-running argument in favor of _not_ having named nodes. It was @dmitrizagidulin who eventually convinced me, but apparently I still don't fully understand all arguments.

timbl:
@retog,  the problem is that the RDF spec, while it may be consistent internally, uses a term "IRI" which is used elsewhere for something different.   The rest of the world uses "IRI" to mean a string, not a node in a graph.  To make our APIs work with developers we need to use terms like IRI (or URI)  in a way consistent wit the way the are used  elsewhere.   
For example I have a  URI library which manipulated URIs (or IRIs) but knows nothing about RDF. The RDF library uses it.

elf-pavlik:
Let's consider example

```ttl
@prefix void: <http://rdfs.org/ns/void#> .
@prefix foaf: <http://xmlns.com/foaf/0.1/> .

<https://x-files.example/>
  a void:DatasetDescription ;
  foaf:primaryTopic <https://x-files.example/#dataset> .
<https://x-files.example/#dataset>
  a void:Dataset ;
  void:uriSpace "https://x-files.example/" .
```

We find here an IRI: `https://x-files.example/` (as defined in [RFC3987](https://tools.ietf.org/html/rfc3987))
> An IRI is a sequence of characters from the Universal Character Set (Unicode/ISO 10646)

Current RDF/JS draft would create two `Term`objects with the same `IRI` as their `value`

```js
{ termType: 'NamedNode',
  value: 'https://x-files.example/' }
{ termType: 'Literal'
  value: 'https://x-files.example/' }
```
One can use common node modules to work with this `IRI` (appearing as `value` of both `NamedNode` and `Literal`)

```node
const url = require('url')
url.parse('https://x-files.example/').protocol
// -> 'https:'
```

IMO clear distinction between `IRI` (string primitive in JS) and `NamedNode`(Term object in RDF/JS), helps with avoiding ambiguity which can lead to confusion. I also think we should add and inline NOTE in section defining `NamedNode` which shortly explains choice of not naming it `IRI`.

retog:
@timbl you could say the same for natural language, most people use the word "word" to refer to a sequence of sounds and not to an element of a sentence. I'm not sure if that's true and people wouldn't mostly expect words and IRIs to carry some meaning but it doesn't matter, what is meant by a word is not its functional role in a sentence nor the set of occurrences in some literature. Analogously in RDF an IRI does not mean a node in a graph it means a resource in the universe, the graph is built by putting IRIs in some functional positions.

@elf-pavlik first I don't see why having "IRI" instead of "NamedNode" as term type would make things harder. Actually I find the use of the value key for the named resource quite confusing, isn't the value of the key the resource (the dataset here) and the IRIs its identifier?

timbl:
I just had a go at looking at what the RDF Concepts and Abstract Syntax document would look like with "Named Node" distinguished from "IRI", and "thing" instead of "resource".  https://www.w3.org/2016/10/rdfcas.html  Not a complete and careful transformation but a quick hack.  I quite like it  though

timbl:
@retog Yes you are right to have "value" for what is actually the name of an Named Node. Is weried.  It is a hack @RubenVerborgh wanted so that code could go faster, as every node type would have a type and value.   I argued in vain that the property should be  .iri instead of .value

RubenVerborgh:
@timbl The main argument was inheritance: it was just elegant to have a property that all terms share, indicating their most important property; `value` suited that purpose. So the value of a named node is its IRI; the value of a blank node its label.

And performance because of uniform object shapes is an added benefit of the single shape.

elf-pavlik:
> in RDF an IRI does not mean a node in a graph it means a resource in the universe
> isn't the value of the key the resource (the dataset here) and the IRIs its identifier

As I understand, IRI *denotes* the resource and multiple IRIs can denote the same resource. The resource itself can have various representations, and IRIs can denote web resources (documents, API endpoints) as well as physical and imaginary resources. Representation of web resource itself can contain serialized RDF graph. Representation of the same resource can contain different serialized graphs at different times ([conversation about g-box & g-snap on gitter](https://gitter.im/rdfjs/public?at=572633514a7c24bd6109eb94)). I remember that we discussed that NamedNode can only have single value, so same resource denoted by two different IRIs would stay represented as two different named nodes. When it comes to values, I think that with exception of Literals, NamedNode and BlankNode act only as references. One needs to dereference to get their (current) value, through the representation of the resource. For non web resources, one can only get a current value of web resource which describes that non web resource, to my understanding we never can get *value* of non web resources.
webarch-uri-resource-representation
*https://www.w3.org/TR/webarch/#p21*

ericprud:
FWIW, my fresh eyes found the uniformity of .value pretty intuitive.
(I'd not followed this until a couple days ago.)
-- 
-ericP

office: +1.617.599.3509
mobile: +33.6.80.80.35.59

(eric@w3.org)
Feel free to forward this message to any list for any purpose other than
email address distribution.

There are subtle nuances encoded in font variation and clever layout
which can only be seen by printing this message on high-clay paper.

interface-spec.md
jeremycarroll: I am somewhat late to this party, but thought I would clarify my intent with the original (which was way back before IRI was a term of art, and we used "RDF URI Reference" as a stand-in). I was aware that for the definition of RDF graph to be mathematically tight, it was necessary for the three sets:  RDF URI References (IRIs); blank nodes; and literals to be disjoint by construction. The definition of blank nodes https://www.w3.org/TR/2004/REC-rdf-concepts-20040210/#section-blank-nodes ensures that they are disjoint. RDF URI references (IRIs) are defined as strings https://www.w3.org/TR/2004/REC-rdf-concepts-20040210/#section-Graph-URIref, whereas literals have named components (which strings do not). https://www.w3.org/TR/2004/REC-rdf-concepts-20040210/#section-Graph-Literal.
Thus all three are disjoint, and I believed at the time that this was sufficiently rigorous and not sloppy. This key design is carried forward into the RDF 1.1 definitions, and the reader who believes them to be sloppy is, IMO, not reading carefully.

jeremycarroll:
I am somewhat late to this party, but thought I would clarify my intent with the original (which was way back before IRI was a term of art, and we used "RDF URI Reference" as a stand-in). I was aware that for the definition of RDF graph to be mathematically tight, it was necessary for the three sets:  RDF URI References (IRIs); blank nodes; and literals to be disjoint by construction. The definition of blank nodes https://www.w3.org/TR/2004/REC-rdf-concepts-20040210/#section-blank-nodes ensures that they are disjoint. RDF URI references (IRIs) are defined as strings https://www.w3.org/TR/2004/REC-rdf-concepts-20040210/#section-Graph-URIref, whereas literals have named components (which strings do not). https://www.w3.org/TR/2004/REC-rdf-concepts-20040210/#section-Graph-Literal.
Thus all three are disjoint, and I believed at the time that this was sufficiently rigorous and not sloppy. This key design is carried forward into the RDF 1.1 definitions, and the reader who believes them to be sloppy is, IMO, not reading carefully.

jeremycarroll:
In terms of @timbl 's rewrite with named node, I think the "referent of a named node" section is the issue. The whole point of literals in the rdf semantics is that they are self-denoting: so a named node labelled with a literal does not refer to anything other than the literal: and the section degenerates into the referent of an IRI.
Of course, introducing an abstract class in an implementation of RDF to mean IRI or string node, vs blank node has its appeal. In Jena we have an abstract class for IRI or blank node. I think when writing the abstract syntax I was loathed to add an additional unnecessary complexity, but left that as an implementation detail that the implementor is free to choose (or not)

PR #95, opened 2016-10-12
Initial High Level spec including Dataset interface

bergos:
This is mainly an entry point for other PR. The `Dataset`interface is a copy of the [wiki page](https://github.com/rdfjs/representation-task-force/wiki/Dataset-spec) without the context, questions and links to RDF-Ext.

### DIFF
--- a/rdfjs-high-level-spec.md
+++ b/rdfjs-high-level-spec.md
@@ -1,0 +1,94 @@
+# RDF/JS High Level API
+
+## Mission Statement
+
+**Status:** This document is a _draft_.<br>
+
+## Data interfaces
+
+### Dataset
+
+**Important:** Unlike Arrays, RDF Datasets have the following properties:
+
+* they are un-ordered (and quads within them cannot be accessed using `[]` notation)
+* like Sets, they are unique (duplicate identical quads are not allowed)
+
+**Properties:**
+
+- `number .length` Number of quads in the dataset. (Similar to `Array.length`) 
+
+**Static/Factory methods:**
+
+- `static Promise<Dataset> import (Function factory, Stream stream)`
+  creates a new Dataset and returns the result of calling `dataset.import(stream)` (see below)
+
+**Array/Iterable compatibility API methods:**
+
+These methods have the same semantics as corresponding methods on an `Array` instance. 
+
+- `Boolean every (Function callback)`
+  tests whether all quads in the dataset pass the test implemented by the provided function
+
+- `Dataset filter (Function callback)`
+  returns a new Dataset with all quads that pass the test implemented by the provided function
+
+- `undefined forEach (Function callback)`
+  executes a provided function once per each quad in the dataset
+
+- `Boolean includes (Quad quad)`
+  determines whether a dataset includes a certain quad, returning `true` or `false` as appropriate
+
+- `Dataset map (Function callback)`
+  returns a new dataset with the results of calling a provided function on every quad in this dataset
+
+- `Boolean some (Function callback)`
+  tests whether some quad in the dataset passes the test implemented by the provided function
+
+**Set-like operations API methods:**
+
+These methods deal with (un-ordered) set operations such as set union, intersection, etc.
+
+- `Dataset add (Quad quad)`
+  adds a given quad to the dataset, returns itself (this dataset)
+
+- `Dataset addAll (Array<Quad>|Dataset quads)`
+  adds all given quads to the dataset, returns itself (this dataset)
+
+- `Dataset difference (Dataset other)`
+  returns a set difference of this dataset and the given one
+
+- `Dataset intersection (Dataset other)`
+  returns a set intersection of this dataset and the given one
+
+- `Dataset merge (Array<Quad>|Dataset other)`
+  returns a set union operation, a new dataset comprised of this dataset merged with all of the quads in the given dataset. (Similar semantics to `Array.concat()`, except that Dataset is of course un-ordered.)
+
+**Misc RDF Dataset API methods:**
+
+- `Dataset clone ()`
+  returns a copy of this dataset (and all its quads)
+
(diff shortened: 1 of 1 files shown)


rdfjs-high-level-spec.md
```diff
@@ -0,0 +1,94 @@
...
+* they are un-ordered (and quads within them cannot be accessed using `[]` notation)
+* like Sets, they are unique (duplicate identical quads are not allowed)
+
+**Properties:**
+
+- `number .length` Number of quads in the dataset. (Similar to `Array.length`) 
+
+**Static/Factory methods:**
+
+- `static Promise<Dataset> import (Function factory, Stream stream)`
```
dmitrizagidulin: What's the `factory` parameter again?

elf-pavlik:
I think we should clarify how Store and Dataset relate to each other. I think sometimes one object might want to have both Store and Dataset interface...

elf-pavlik:
This gitter comment brought to my attention one of the distinctions between Dataset and Store https://gitter.im/linkeddata/rdflib.js?at=580e26dc684ccefd38a081fa

Stores could do indexing, but Dataset would just provide more lightweight inteface for a set of quads, in many cases more convenient to use than a raw stream of quads.

RubenVerborgh:
Isn't a Store an implementation of a Dataset then?

elf-pavlik:
As I understood @dmitrizagidulin rdflib.js would not implement distinct Store and Dataset.

rdfjs-high-level-spec.md
elf-pavlik: I think it forwards data factory which dataset should use https://github.com/rdfjs/representation-task-force/issues/79

jacoscaz:
I think some sort of easy-to-use abstraction for dealing with sets of quads being passed around between functions is necessary. I find that one of the hardest thing to do in RDF land is creating the equivalent of simple objects like `{power: true, dimming: 7}` in a straight-forward way. 

We experimented with (de)serialising RDF data to instances of a custom wrapper around N3's Store that adds chain-based features like 

```
set.subject('<http://example.com/s>')
  .predicate('<http://example.com/power>', true)
  .predicate('<http://example.com/dimming>', 7)
```

but felt that it still becomes too complex. We're now looking at a more descriptive approach via something like `simplerdf` + RDF/JS interface.

In any case, I think the API between Dataset and Store should be kept as different as possible to make life easier for implementors.

RubenVerborgh:
@jacoscaz Indeed, simplerdf was actually the final trigger to get the RDF/JS work started. If all low-level libraries implement this spec, then simplerdf can be compatible with all of them.

dmitrizagidulin:
> In any case, I think the API between Dataset and Store should be kept as different as possible to make life easier for implementors.

Can you explain a bit more?

elf-pavlik:
Besides conversation in https://github.com/rdf-ext/discussions/issues/21
> @bergos: The difference between a Store and Graph is more the sync vs. async interface.

We also had some discussion with @bergos during TF calls about possible requirement for Store interface anticipate using it for stores that expose interface to remote servers (LDP, SPARQL, Triple Patter Fragmetns ...) Which will have different requirements to Dataset which to my understanding should just represent set of quads available in memory.

blake-regalia:
@jacoscaz 
> I think some sort of easy-to-use abstraction for dealing with sets of quads being passed around between functions is necessary.

This is where libraries can take their own approach to simplifying the process, such as with creating/inserting triples. 

> I find that one of the hardest thing to do in RDF land is creating the equivalent of simple objects

[Like this](https://github.com/blake-regalia/graphy.js/blob/serializer-example/examples/serialize-ttl.js)? The output is [shown here](https://github.com/blake-regalia/graphy.js/blob/serializer-example/examples/output.ttl). I'll be releasing this w/ next version of graphy.

niklasl:
@jacoscaz @blake-regalia Without specific application needs I cannot really judge what's required, but why not just JSON-LD? Like:

```json
{
  "@context": {"@vocab": "http://example.com/"},
  "@id": "http://example.com/s",
  "power": true,
  "dimming": 7
}
```

bergos:
`NamedNode`, `BlankNode`, `Literal` and `Quad` are all in memory representations of the RDF model. Dataset completes the RDF model. It's aligned to the JS Array API. It's not designed to connect to external stores like the `Store` interface. It's designed to pass multiple quads in JS code and sync access. But I agree that the label `high level` is confusing.

I would like to shift the discussion, if and how we add it to the spec, to a later time and focus now on the stream interfaces. Of course everybody who want's to use or implement the API is invited to discuss technical details.

This is not about about an application specific high level API like SimpleRDF or other mentioned libraries. We can make a wiki page for it and an issue to discuss it, but let's not mix topics here.

jacoscaz:
@niklasl @blake-regalia yes, we're also experimenting with pre-compacting JSON-LD documents down to plain old JSON-objects using shared contexts before when passing data from RDF-aware components to non-RDF-aware components.

l00mi:
How do we want to proceed here?
Thumbs Up: Merge and open multiple Issues/PR to adopt the initial proposition.
Thumbs Down: Close the request and start off with a more basic version.

elf-pavlik:
If we decide to merge it let's create right away an Issue (or PR if someone feels up to it) which clarifies similarities and differences between Store and Dataset

l00mi:
Can please everybody cast your vote to go with this proposition as initial draft for the high-level spec?

This will lay the ground for the future topics discussed in https://github.com/rdfjs/representation-task-force/issues/123, https://github.com/rdfjs/representation-task-force/issues/108, https://github.com/rdfjs/representation-task-force/issues/100, https://github.com/rdfjs/representation-task-force/issues/98.

bergos:
In favor of the new [Dataset Spec](https://github.com/rdfjs/dataset-spec), I would like to close this PR.

PR #99, opened 2016-10-21
remove Store#deleteGraph

elf-pavlik:
`store.removeMatches(null, null, null, term)`already covers that

### DIFF
--- a/interface-spec.md
+++ b/interface-spec.md
@@ -200,7 +200,3 @@ Access to stores LDP or SPARQL endpoints can be implemented with a Store intefac
 - `EventEmitter .removeMatches([Term|RegExp subject], [Term|RegExp predicate], [Term|RegExp object], [Term|RegExp graph])`
   All quads matching the pattern will be removed.
   The `end` and `error` events are used like described in the `Stream` interface.
-
-- `EventEmitter .deleteGraph(Term|string graph)`
-  Deletes the given named graph.
-  The `end` and `error` events are used like described in the `Stream` interface.


bergos:
`.removeMatches` just deletes the triples/quads. The result would be an empty named graph. If you would access it via HTTP you would get a 204 response. After calling `deleteGraph` you would get a 404.

elf-pavlik:
https://www.w3.org/TR/rdf11-concepts/#section-dataset
> NOTE
Some RDF dataset implementations do not track empty named graphs. Applications can avoid interoperability issues by not ascribing importance to the presence or absence of empty named graphs.

But actually [Web Access Control](https://github.com/solid/web-access-control-spec) pretty much sets access on named graphs. So empty named graphs can still have custom ACL. So an agent with write access could start adding triples to it (but not quads).

I think we should add a note to the spec about empty named graphs and difference between
* `store.removeMatches(null, null, null, term)`
* `store.deleteGraph(term)`

RubenVerborgh:
The discussion here should not involve HTTP interfaces IMHO.

blake-regalia:
I think there are use cases for triple stores that would prefer to distinguish between truncation and deletion for the purpose of conserving or reusing a graph container.

bergos:
Agree. Please vote -1 to close this PR without merging.

elf-pavlik:
Looking at https://github.com/rdfjs/representation-task-force/blob/master/interface-spec.md#source
when using `.match()` empty named graph or no graph doesn't seem to make difference.

Requirement to have both

* `Store#removeMatches(null, null, null, term)`
* `Store#deleteGraph(term)`

Might require providing little more guidance about the difference (since it doesn't seem to make one for results of `Store#match`. Also referencing https://www.w3.org/TR/rdf11-concepts/#section-dataset might help implementers to stay informed
> NOTE
Some RDF dataset implementations do not track empty named graphs. Applications can avoid interoperability issues by not ascribing importance to the presence or absence of empty named graphs.

`

l00mi:
@elf-pavlik are you happy to convert this into an Issue to improve the documentation?

elf-pavlik:
I think we can just document well the difference between Store and Dataset proposed in #95 explaining that Store anticipates for remote stores thus methods like `deleteGraph`

Issue #106, opened 2017-03-15
The Sink .import() method doesn't provide feedback


jacoscaz:
The `.import(stream.Readable stream)` method from the `Sink` interface does not seem to provide any feedback regarding the state of the import operation as it does not return an `events.EventEmitter` and/or take a callback as an argument.

In `node-quadstore` I've opted to have it return an instance of `events.EventEmitter` to make it consistent with `remove()`. Should I make a PR for this?

elf-pavlik:
Looking at your example https://github.com/beautifulinteractions/node-quadstore#rdfstoreprototypeimport

```js
const readableStream; // A stream.Readable of Quad() instances

store.import(readableStream)
  .on('error', (err) => {})
  .on('end', () => {});
```
I think currently one needs to handle events on that `readableStream` passed to `import`

```js
const readableStream // A stream.Readable of Quad() instances
  .on('error', (err) => {})
  .on('end', () => {})

store.import(readableStream)
```

@RubenVerborgh have already implemented `import` interface in N3.js ?

RubenVerborgh:
Hmm, maybe it's not because the input stream has been read, that all triples have been processed.

Haven't implemented that yet in N3.js, I'm really behind. Hoping for April.

jacoscaz:
I agree with @RubenVerborgh. The `end` event should signal the actual termination of the import procedure, which might happen some time after the source stream has ended. The `error` event, on the other hand, might _also_ relay errors from the source stream, although I'm not 100% sure about that.

elf-pavlik:
> The end event should signal the actual termination of the import procedure, which might happen some time after the source stream has ended.

Makes sense, @jacoscaz would like to make another PR for it?

bergos:
I had the same problem of the missing feedback and implemented it with an EventEmitter: https://github.com/rdf-ext/rdf-sink/blob/develop/index.js#L9

elf-pavlik:
addressed in #110

PR #110, opened 2017-03-16
Changes the return type of Sink .import() from undefined to EventEmitter

jacoscaz:
Takes care of #106

### DIFF
--- a/interface-spec.md
+++ b/interface-spec.md
@@ -185,7 +185,7 @@ A Sink is an object that consumes quads.
 It can store the quads or do some further processing.
 For example serializers and transformations which consume quads can implement the Sink interface.
 
-- `undefined .import(Stream stream)`
+- `EventEmitter .import(Stream stream)`
   Writes all quads from the stream to the sink.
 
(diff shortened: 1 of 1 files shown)


elf-pavlik:
Should we clarify that implementation SHOULD NOT use 'data' events on it but only 'end' and 'error'? Maybe even suggesting if it SHOULD/MUST forward errors from the stream passed to import or not...

jacoscaz:
I think we should but I would like to know more about the reason why `events.EventEmitter` was selected for `.remove()` (as opposed to a callback or a Promise) before making up my mind. Not that I do not like it!

bergos:
`EventEmitter` was choosen for consistency reasons. In `rdf-ext` the streams can be wrapped into Promises by using the `Dataset` like this:
```
rdf.dataset().import(quadStream).then((dataset) => {
  ...
})
```

For events `rdf.waitFor` can be used:
```
rdf.waitFor(store.removeMatches(iri)).then(() => {
  ...
})
```

PR #113, opened 2017-04-09
generic Sink.import for Parser, Serializer, Transform and Store

bergos:
Like I described already in [this comment](https://github.com/rdfjs/representation-task-force/pull/80#issuecomment-289426681), only the `.import` method can be used for all use cases. The develop branches of rdf-ext are already implemented like this. It works very well and the code looks also very readable.

### DIFF
--- a/interface-spec.md
+++ b/interface-spec.md
@@ -181,12 +181,24 @@ For example, parsers and transformations which generate quads can implement the 
 
 ### Sink
 
-A Sink is an object that consumes quads.
-It can store the quads or do some further processing.
-For example serializers and transformations which consume quads can implement the Sink interface.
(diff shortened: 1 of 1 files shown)


interface-spec.md
```diff
@@ -181,12 +181,24 @@ For example, parsers and transformations which generate quads can implement the
...
+  Depending on the use case, subtypes of `EventEmitter` or `Stream` are used.
+
+  Typical use cases:
+
+  **Parser:** `Source .import(Stream stream)`
+
+  **Serializer:** `Stream .import(Source stream)`
+
+  **Transformation:** `Source .import(Source stream)`
+
```
bergos: @elf-pavlik I've updated the PR.

Issue #130, opened 2018-11-21
Evaluating the creation of PHP classes and interfaces based on the RDF.js specification


k00ni:
I am one of the maintainers/supporters of RDF libraries such as hardf, ARC2 and Saft, written in PHP. And i am currently evaluating, if its worth working on a (basic) PHP class set, which reflects the latest state of your standard/specification. It would be great if you could answer some of my questions:

1. Is this the latest version of the specification? http://rdf.js.org/
2. Where can i get further details about the usage/meaning of certain interfaces/classes? *What i found so far, the wiki here does not contain sufficient information about that topic.*
3. Is there any activity in this task force recently? Whats the plan for 2019?

Thanks in advance!

**Internal references:**
* #81
* #124
* #131
* #132
* #133
* #134 (pull request)

**External references:**
* Providing a draft to create a new RDF specification for PHP libraries (https://github.com/php-fig/fig-standards/pull/958)

k00ni:
ping @BigBlueHat

bergos:
1. Yes, that's the latest version
2. There are some up-to-date packages in this org and rdf-ext which mention the used interface (e.g. https://github.com/rdfjs/serializer-ntriples, https://github.com/rdf-ext/rdf-store-sparql). Also we are working on the documentation of rdfjs/rdf-ext. It's still work in progress: https://github.com/rdf-ext/documentation. Does this help? Otherwise, please ask more detailed questions. This would also help to shape the spec and documentation.
3. There was not much activity around the spec itself, but more implementations are using it. We didn't discuss a schedule, but I think beginning of next year would be a good time to finalize the spec.

elf-pavlik:
#67 also have list of known implementations, as well as work in progress.

k00ni:
If someone is interested in the current state, here is the repository:

https://github.com/rdfphp/baseline

elf-pavlik:
Very cool :exclamation: 

> If possible, constructors are used to set the value of the instance. For instance `$node = new NamedNode('foo:bar')`;

I think in low level interfaces we defined so far we require absolute IRIs. How does your interface handle CURIEs

bergos:
We don't define any IRI validation. This allows also relative paths and CURIEs. That's on purpose, so libraries can implement features like resolving CURIEs. In general, some definitions are very open to allow libraries to add new features and still be valid according to the spec.

k00ni:
@elf-pavlik wrote:
[...]

PR #134, opened 2018-11-28
Add info to §1. Design elements and principles, that specification does not define any value validation

k00ni:
## Why?

For some people it may be obvious, but if you not that familiar with the rdfjs specification, you may wonder if an implementation should implement *value validation* or not.

@bergos stated it [here](https://github.com/rdfjs/representation-task-force/issues/130#issuecomment-442227907) and for me as an implementer its very important to know that.

## What?

I added a point in `§1. Design elements and principles`, which says that specification does not define any validation. If you want to change the text, feel free.

*Ref https://github.com/rdfjs/representation-task-force/issues/130*

### DIFF
--- a/interface-spec.html
+++ b/interface-spec.html
@@ -108,7 +108,7 @@
     <li>Interfaces may have additional implementation specific properties.</li>
     <li>
        We don't define any validation of given values (e.g. IRI, URI, CURIE). 
-       If an implementation enforces valid values, it should state details prominent for the user.
+       Implementations that apply validation should make this fact clear in their documentation.
     </li>
   </ul>
 
--- a/interface-spec.html
+++ b/interface-spec.html
@@ -108,7 +108,8 @@
     <li>Interfaces may have additional implementation specific properties.</li>
     <li>
        We don't define any validation of given values (e.g. IRI, URI, CURIE). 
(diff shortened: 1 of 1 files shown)


interface-spec.html
```diff
@@ -106,6 +106,10 @@ <h2>Design elements and principles</h2>
     </li>
     <li>Should allow "upgrading" a plain object into a fully functional triple</li>
     <li>Interfaces may have additional implementation specific properties.</li>
+    <li>
+       We don't define any validation of given values (e.g. IRI, URI, CURIE). 
+       If an implementation enforces valid values, it should state details prominent for the user.
```
vhf: @k00ni I'm not sure to understand this part:

> it should state details prominent for the user

Could you please explain and we'll find a better wording?

vhf: Thanks, I get the idea now!

```suggestion
       Implementations that apply validation should make this fact clear in their documentation.
```

Something like this?

k00ni: :+1:

k00ni:
I have committed your suggestion @vhf.

k00ni:
Ping :)

elf-pavlik:
Due to conflicts I needed to pull your commits manually to master branch, thanks!

k00ni:
No problem.

Issue #128, opened 2018-09-26
Store matching RegExps


blake-regalia:
What is the expectation here for the underlying string that the regexps are being matched on? The reasonable thing that comes to mind would be to clarify in the spec that implementations must match the regex against some canonical string for a term such as N-Triple, but this opens up questions. Are blank node labels be preserved? Should the default graph be ignored if a RegExp is used?

bergos:
I would propose to remove `RegExp` from the interface. @jacoscaz proposed some extensions in #123 . We could move the ideas from the other issue and the `RegExp` stuff into a advanded filtering and/or let implementors move ahead with a proposal.

rubensworks:
> I would propose to remove RegExp from the interface.

I strongly agree with this.

The `RexExp` params complicate the implementation of `Store` too much. While it may be useful for consumers, it does not really belong there IMO.

RubenVerborgh:
Yes please.

PR #140, opened 2019-01-21
removed RegExp & added nullable to match methods

### LINKED ISSUE #139: does match considers null as 'optional' parameter value?
I've just tried using TypeScript definitions from `@types/rdf-js` by @rubensworks and @LaurensRietveld which [include this line](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/e9038e51b914f6a0da5cb559956bfaa35db84735/types/rdf-js/index.d.ts#L356)

```ts
match(subject?: Term | RegExp, predicate?: Term | RegExp, object?: Term | RegExp, graph?: Term | RegExp): Stream<Q>
```

VScode gives me a hint that I can use `undefined` but it seams I can't use `null`, I think current wording in the spec doesn't make it clear: http://rdf.js.org/#source-interface

```webidl
Stream match(optional (Term or RegExp) subject, optional (Term or RegExp) predicate, optional (Term or RegExp) object, optional (Term or RegExp) graph);
```

I thought that since we consider removing `RegExp` from that interface, we could also clarify handling of `null`.

bergos:
would fix #139 and #128.

see [PR in Dataset spec](https://github.com/rdfjs/dataset-spec/pull/38#issue-246313067) for more details about nullable.

### DIFF
--- a/interface-spec.html
+++ b/interface-spec.html
@@ -442,7 +442,7 @@
 
     <pre class="idl">
     interface Source {
-      Stream match(optional (Term or RegExp) subject, optional (Term or RegExp) predicate, optional (Term or RegExp) object, optional (Term or RegExp) graph);
+      Stream match(optional Term? subject, optional Term? predicate, optional Term? object, optional Term? graph);
     };
     </pre>
 
@@ -490,7 +490,7 @@
     <pre class="idl">
     interface Store {
       EventEmitter remove(Stream stream);
-      EventEmitter removeMatches(optional (Term or RegExp) subject, optional (Term or RegExp) predicate, optional (Term or RegExp) object, optional (Term or RegExp) graph);
+      EventEmitter removeMatches(optional Term? subject, optional Term? predicate, optional Term? object, optional Term? graph);
       EventEmitter deleteGraph((Term or string) graph);
     };
 


elf-pavlik:
Awesome! I don't know if we should just rely on people implementing it to understand WebIDL `?` or we could add one line note clarifying it and linking to specific definition in WebIDL spec.

rubensworks:
I assume this will be part of a major version update to the spec? As this is a breaking change.

elf-pavlik:
only 'removed RegExp' looks breaking, do we know of anyone who have implemented it?

rubensworks:
During the call it was mentioned that the spec is still in a stage that allows breaking changes, so I retract my last comment. In any case, it is important to explicitly mention this change somewhere though.

> only 'removed RegExp' looks breaking, do we know of anyone who have implemented it?

As far as I know, only Store implementations exist that implement it, but no one actually calls Stores with RegExps. So if that is true, no existing applications should break.

bergos:
We never defined the exact handling of RegExp parameters see #128. Removing it now is a breaking change from the perspective of code that's using the feature. From library perspective, RegExp support is now a custom feature.

- For rdf-ext I can rule out that people used it, as it wasn't supported yet.
- @jacoscaz I only had a quick look at quadstore, but I didn't see RegExp support. So I think that would be also not problem for your code.
- I'm not aware of any other implementation which could be affected by this change.

Issue #137, opened 2019-01-15
Add DataFactory method to create from existing term


RubenVerborgh:
We want to consider adding a method to `DataFactory` that takes a `Term`-shaped object (from _any_ library) as input, and outputs a `Term` (from the library of the `DataFactory`) that `equals` it.

This can be used to convert a `Term` from one RDF/JS implementation to another.

rubensworks:
For consistency, we may also want add a similar method for converting `Quad` objects.

elf-pavlik:
I could make PR for `DataFactory#clone()`

```webidl
(Term or Quad) clone((Term or Quad) other)
```

I think in this case we don't allow `undefined` or `null`

rubensworks:
@elf-pavlik I would propose creating separate methods for cloning terms and quads (`cloneTerm`, `cloneQuad`). Implementations may otherwise become too complicated for determining the argument type.

RubenVerborgh:
fromTerm/fromQuad?

bergos:
👎  for `.clone()`, as this could be interpreted as a clone function of the factory itself. It doesn't make sense now, but maybe in the future we have an additional spec (Prefix + CURIE support which turns the DataFactory into an environment) were having multiple instance of the DataFactory makes more sense.

I favor `fromTerm` and `fromQuad` a little bit over `cloneTerm` and `cloneQuad`. `from*` indicates much stronger that new instances by the DataFactory are created and not that somehow the constructor of the given Term/Quad is used.

elf-pavlik:
`fromTerm` & `fromQuad` SGTM, if no one else makes PR today/tomorrow I'll submit one on Wednesday

elf-pavlik:
PR #158

awwright:
Sorry I just noticed this, but I'm curious... Why not pass a plain object to the Term or Quad functions directly?

elf-pavlik:
As I replied on https://gitter.im/rdfjs/public?at=5cb48398c55bd14d35ebcf88 single factory object makes it much easier to pass it around. Spec only defines methods on the factory and doesn't define any constructors for specific terms.

Issue #132, opened 2018-11-24
Usage of exceptions/errors?


k00ni:
The specification currently doesn't specific certain errors or exceptions. Is that intended? Would you open to define ones for certain cases?

One would be if someone tries to change an instance of `Term` or `Quad`. But perhaps the decision was already made against that (ref #81).

Another one, if you use the wrong `Term` subclass in a quad, e.g. predicate = BlankNode. 

(ref #130)

bergos:
Both cases are not handled by the spec. Also in JS it's not very common to make custom error classes. And `instanceof` should not be used, as this could cause problems with different implementations.

blake-regalia:
I think this is important for factory methods as well. For instance, if `null` is passed to `NamedNode.equals()`, should this throw an Error or return false? Right now, the [org data model test suite](https://github.com/rdfjs/data-model/commit/417efb2cfc1013e6c4f61818248264afe562fc5b) requires these methods return false but this is not defined in the interface specification -- I recommended we either remove this from the test suite or define it in the interface spec.

k00ni:
I'd rather prefer clear exceptions or just errors, but with defined error texts, instead of returning false. In a complex environment, a returned false says nothing about the state of the previous operation. Giving the user a way to gather critical information is important, if you handle invalid data, while parsing for instance.

k00ni:
As a starting point, i propose the following exceptions:

1. **InvalidTermUsageInQuadException**
* Gets thrown, if you use an invalid instance of `Term` in `Quad`.
* Example: use a `Literal` instance as subject

2. **ObjectNotMutableException**
* Gets thrown, if you try to change an instance of `Term` or `Quad`.

3. **UndefinedPropertyException**
* Gets thrown, if you try to access an undefined property of an instance of `Term` or `Quad`.

RubenVerborgh:
2 is not idiomatic; 3 not enforceable on regular objects (except for proxies)

blake-regalia:
I think the best we can hope for would be to say that certain methods should throw an Error, maybe even for example a [TypeError](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypeError), if provided an argument that is not a Term, Quad, etc. This is critical for debugging if a variable in a dev's code is not properly initialized and passes `undefined` to say, `.equals(term)`. Returning false instead of throwing is not a convenience, it only further obscures the source of such problems.

blake-regalia:
[...]

Issue #104, opened 2017-03-08
Nuance "plain objects (JSON) cannot be used."


RubenVerborgh:
The statement

> Given the necessity of methods, plain objects (JSON) cannot be used.

if confusing; we probably want something more correct.

elf-pavlik:
This seems relevant to https://github.com/rdfjs/representation-task-force/issues/79 (forwarding DataFactory). One could pass factory that creates objects without methods?

RubenVerborgh:
That would be a spec-incompatible factory, but yes, could be so for simplicity.

elf-pavlik:
This issue in rdflib.js by @fletcher91 seems relevant: https://github.com/linkeddata/rdflib.js/issues/219 *(Functional-style datamodel & performance)*

vhf:
While we are at it, I'd suggest removing `(JSON)` because JSON only exists as a string, not as object(s).

I understand that what is meant by "JSON objects" is something like _objects created using the literal syntax or `Object.create(null)`_ but I still find the phrase _plain objects (JSON)_ confusing.

elf-pavlik:
If we want to ship at least the Data interfaces spec, do we still need to address something in this issue?

> This issue in rdflib.js by @fletcher91 seems relevant: linkeddata/rdflib.js#219 (Functional-style datamodel & performance)

@RubenVerborgh I think currently you might have the best understanding of rdflib issues, I see that issue mentioned above resulted in PR which claims: https://github.com/linkeddata/rdflib.js/pull/272#issuecomment-437709049

> So, I've created a turtle parsing benchmark, and the results show about a 40% decrease in memory usage (~55% after garbage collection), and an 8% reduction in processing time.

> Looking at the difference before and after the gc call, there seems to be some additional space for reducing memory usage.

RubenVerborgh:
> In fact, any library should be able to convert plain JSON objects into objects with methods.

Relevant: #137

RubenVerborgh:
> > This issue in rdflib.js by @fletcher91 seems relevant: [linkeddata/rdflib.js#219](https://github.com/linkeddata/rdflib.js/issues/219) (Functional-style datamodel & performance)
> 
> @RubenVerborgh I think currently you might have the best understanding of rdflib issues, I see that issue mentioned above resulted in PR which claims: [linkeddata/rdflib.js#272 (comment)](https://github.com/linkeddata/rdflib.js/pull/272#issuecomment-437709049)

[...]

PR #142, opened 2019-01-21
Rigorously define equals method on Term and Quad

### LINKED ISSUE #133: Quad interface: when does equals return true?
The spec says:

> equals() returns true if and only if the argument is a) of the same type b) has all components equal.

The only parameter is of type `Quad`, therefore (a) is always right, isn't it? Or what is meant with "of the same type"?

If my assumption is right, i would suggest removing the (a) part from the text.

(ref #130)

RubenVerborgh:
Closes #133.

### DIFF
--- a/index.html
+++ b/index.html
@@ -131,7 +131,7 @@
     interface Term {
       attribute string termType;
       attribute string value;
-      boolean equals(Term other);
+      boolean equals(optional Term? other);
     };
     </pre>
 
@@ -170,7 +170,7 @@
     interface NamedNode : Term {
       attribute string termType;
       attribute string value;
-      boolean equals(Term other);
+      boolean equals(optional Term? other);
     };
     </pre>
 
@@ -195,7 +195,7 @@
     interface BlankNode : Term {
       attribute string termType;
       attribute string value;
-      boolean equals(Term other);
+      boolean equals(optional Term? other);
     };
     </pre>
 
@@ -224,7 +224,7 @@
       attribute string value;
       attribute string language;
       attribute NamedNode datatype;
-      boolean equals(Term other);
+      boolean equals(optional Term? other);
     };
     </pre>
 
@@ -265,7 +265,7 @@
     interface Variable : Term {
       attribute string termType;
       attribute string value;
-      boolean equals(Term other);
+      boolean equals(optional Term? other);
     };
     </pre>
 
@@ -291,7 +291,7 @@
     interface DefaultGraph : Term {
       attribute string termType;
       attribute string value;
-      boolean equals(Term other);
+      boolean equals(optional Term? other);
     };
     </pre>
 
@@ -321,7 +321,7 @@
       attribute Term predicate;
       attribute Term object;
       attribute Term graph;
-      boolean equals(Quad other);
+      boolean equals(optional Quad? other);
     };
     </pre>
 
--- a/index.html
+++ b/index.html
@@ -345,8 +345,20 @@
       <code>Triple</code> MUST be represented as <code>Quad</code> with <code>graph</code> set to a <code>DefaultGraph</code>
     </p>
     <p>
-      <dfn>equals()</dfn> returns <code>true</code> if and only if the argument is a) of the same
-      type b) has all components equal.
+      <dfn>equals()</dfn> returns <code>true</code>
+      when called with parameter <code>other</code>
+      on an object <code>quad</code> if
+      all of the conditions below hold:
+    </p>
+    <ul>
+      <li><code>other</code> is <em>neither</em> <code>null</code> nor <code>undefined</code>;</li>
+      <li><code>quad.subject.equals(other.subject)</code> evaluates to <code>true</code>;</li>
+      <li><code>quad.predicate.equals(other.predicate)</code> evaluates to <code>true</code>;</li>
+      <li><code>quad.object.equals(other.object)</code> evaluates to a <code>true</code>;</li>
+      <li><code>quad.graph.equals(other.graph)</code> evaluates to a <code>true</code>;</li>
+    </ul>
+    <p>
+      otherwise, it returns <code>false</code>.
     </p>
   </section>
 
--- a/index.html
+++ b/index.html
@@ -148,8 +148,18 @@
       <dfn>value</dfn> is refined by each interface which extends Term.
     </p>
     <p>
-      <dfn>equals()</dfn> returns <code>true</code> if and only if <code>other</code> has the same
-      <code>termType</code> and the same contents (as defined by concrete subclasses).
+      <dfn>equals()</dfn> returns <code>true</code>
+      when called with parameter <code>other</code>
+      on an object <code>term</code> if all of the conditions below hold:
+    </p>
+    <ul>
+      <li><code>other</code> is <em>neither</em> <code>null</code> nor <code>undefined</code>;</li>
+      <li><code>term.termType</code> is the same string as <code>other.termType</code>;</li>
+      <li><code>other</code> follows the additional constraints of the specific <code>Term</code> interface implemented by <code>term</code>
+        (e.g., <a>NamedNode</a>, <a>Literal</a>, …);</li>
+    </ul>
+    <p>
+      otherwise, it returns <code>false</code>.
     </p>
   </section>
 
@@ -171,8 +181,10 @@
       <dfn>value</dfn> the IRI of the named node (example: <code>"http://example.org/resource"</code>).
     </p>
     <p>
-      <dfn>equals()</dfn> returns <code>true</code> if and only if <code>other</code> has
-      <code>termType</code> <code>"NamedNode"</code> and the same <code></code>value</code>.
+      <dfn>equals()</dfn> returns <code>true</code> if
+      all general <a>Term.equals</a> conditions hold
+      and <code>term.value</code> is the same string as <code>other.value</code>;
+      otherwise, it returns <code>false</code>.
     </p>
   </section>
 
@@ -196,8 +208,10 @@
       sourced from RDF/XML, do not change the blank node name (example: <code>"blank3"</code>)
     </p>
     <p>
-      <dfn>equals()</dfn> returns <code>true</code> if and only if <code>other</code> has
-      <code>termType</code> <code>"BlankNode"</code> and the same <code>value</code>.
+      <dfn>equals()</dfn> returns <code>true</code> if
+      all general <a>Term.equals</a> conditions hold
+      and <code>term.value</code> is the same string as <code>other.value</code>;
+      otherwise, it returns <code>false</code>.
     </p>
   </section>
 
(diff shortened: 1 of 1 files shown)


interface-spec.html
```diff
@@ -146,8 +146,18 @@ <h3><dfn>Term</dfn> interface</h3>
...
-      <code>termType</code> and the same contents (as defined by concrete subclasses).
+      <dfn id="term-equals">equals()</dfn> returns <code>true</code>
+      when called with parameter <code>other</code>
+      on an object <code>term</code> if and only if
+      all of the conditions below hold:
+    </p>
+    <ul>
+      <li><code>other</code> is <em>neither</em> <code>null</code> nor <code>undefined</code>;</li>
+      <li><code>term.termType</code> is the same string as <code>other.termType</code>;</li>
+      <li><code>other</code> follows the additional constraints followed by the most specific <code>Term</code> interface implemented by <code>term</code> (as detailed below);</li>
```
rubensworks: `followed by the most specific <code>Term</code>`

This text makes it seem that multiple interfaces could be implemented, which shouldn't be possible. Should this be _"the more specific Term"_ instead?

interface-spec.html
```diff
@@ -233,9 +247,12 @@ <h3><dfn>Literal</dfn> interface</h3>
...
     </p>
     <p>
-      <dfn>equals()</dfn> returns <code>true</code> if and only if <code>other</code> has
-      <code>termType</code> <code>"Literal"</code> and the same <code>value</code>,
-      <code>language</code>, and <code>datatype</code>.
+      <dfn>equals()</dfn> returns <code>true</code> if and only if
+      <a href="#term-equals">all general <code>Term#equals</code> conditions</a> hold,
+      <code>term.value</code> is the same string as <code>other.value</code>,
+      <code>term.language</code> is the same string as <code>other.language</code>, and
+      <code>term.dataType.equals(other.dataType)</code> evaluates to a truthy value;
```
rubensworks: `<code>term.datatype.equals(other.datatype)</code>`

interface-spec.html
blake-regalia: I must admit that I don't understand what this last item means about the more specific Term interface?

interface-spec.html
```diff
@@ -334,8 +354,20 @@ <h3><dfn>Quad</dfn> interface</h3>
...
+      when called with parameter <code>other</code>
+      on an object <code>quad</code> if and only if
+      all of the conditions below hold:
+    </p>
+    <ul>
+      <li><code>other</code> is <em>neither</em> <code>null</code> nor <code>undefined</code>;</li>
+      <li><code>quad.subject.equals(other.subject)</code> evaluates to a truthy value;</li>
+      <li><code>quad.predicate.equals(other.predicate)</code> evaluates to a truthy value;</li>
+      <li><code>quad.object.equals(other.object)</code> evaluates to a truthy value;</li>
+      <li><code>quad.graph.equals(other.graph)</code> evaluates to a truthy value;</li>
```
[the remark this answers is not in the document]
RubenVerborgh: The idea was to relax, so you don’t have to write `=== true`; you might not own the implementations of the component methods (terms could be imported).

interface-spec.html
RubenVerborgh: It means NamedNode/Literal/… will rephrase, probably enumerate.

interface-spec.html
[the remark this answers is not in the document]
blake-regalia: Okay, so then we also have to check that `quad.[subject/predicate/object/graph].equals` are indeed functions before using them yes? Otherwise they will throw an error.

RubenVerborgh:
Checking for function is extra work; checking for truthy instead of true is less. I don’t mind though; let me know if you want true and I’ll change, I just don’t think it’s necessary to go the extra mile.

blake-regalia:
Yes, but isn't the idea to always return `true` or `false` and not throw an error? What if i pass a term to quad constructor such that `quad.term.equals === null`? We need to test at runtime that `.equals` is a function for each component of quad; and that this is another consequence of #132 / #133

RubenVerborgh:
I don’t feel strongly, will just write true.

interface-spec.html
```diff
@@ -146,12 +146,24 @@ <h3><dfn>Term</dfn> interface</h3>
...
+      <dfn id="term-equals">equals()</dfn> returns <code>true</code>
+      when called with parameter <code>other</code>
+      on an object <code>term</code> if and only if
+      all of the conditions below hold:
+    </p>
+    <ul>
+      <li><code>other</code> is <em>neither</em> <code>null</code> nor <code>undefined</code>;</li>
+      <li><code>term.termType</code> is the same string as <code>other.termType</code>;</li>
+      <li><code>other</code> follows the additional constraints of the specific <code>Term</code> interface implemented by <code>term</code>
+        (e.g., <a href="#namednode-interface"><code>NamedNode</code></a>,
```
bergos: Just `<a>NamedNode</a>` is enough to link internal. Respec detects that it's an interface and will format it like `code`.

interface-spec.html
```diff
@@ -146,12 +146,24 @@ <h3><dfn>Term</dfn> interface</h3>
...
+        <a href="#literal-interface"><code>Literal</code></a>, …);</li>
+    </ul>
+    <p>
+      otherwise, it returns <code>false</code>.
     </p>
   </section>
 
-  <section data-dfn-for="NamedNode">
+  <section id="namednode-interface" data-dfn-for="NamedNode">
```
bergos: Using the simple version of `<a></a>` doesn't require any `id` attributes.

interface-spec.html
```diff
@@ -169,12 +181,14 @@ <h3><dfn>NamedNode</dfn> interface</h3>
       <dfn>value</dfn> the IRI of the named node (example: <code>"http://example.org/resource"</code>).
     </p>
     <p>
-      <dfn>equals()</dfn> returns <code>true</code> if and only if <code>other</code> has
-      <code>termType</code> <code>"NamedNode"</code> and the same <code></code>value</code>.
+      <dfn>equals()</dfn> returns <code>true</code> if and only if
+      <a href="#term-equals">all general <code>Term#equals</code> conditions</a> hold
```
bergos: Respec would also handle this case if `<a>Term.equals</a>` is used. As the link would not cover the current text, maybe change it to:

`if and only if all general condititions defined in <a>Term.equals</a>`

blake-regalia:
There is still an outstanding edge case that is not covered by the algorithm which I brought up in my last comment; namely that `.equals` needs to be tested for otherwise what happens when:

```js
factory.quad({
	termType: 'NamedNode',
	value: 'http://ex.org/s',
}, {
	termType: 'NamedNode',
	value: 'http://ex.org/p',
}, {
	termType: 'NamedNode',
	value: 'http://ex.org/o',
}, {
	termType: 'NamedNode',
	value: 'http://ex.org/g',
}).equals(other);
```

will throw: `Uncaught TypeError: {...}.equals is not a function`

RubenVerborgh:
@blake-regalia We are assuming now that they are valid `Term`s; I can state that assumption somewhere explicitly, but I'd prefer to keep it. Too complex and expensive otherwise.

Also see #137.

interface-spec.html
RubenVerborgh: Mmm, don't personally like this, but WebIDL has to be followed, thanks for noticing.

RubenVerborgh:
@bergos' comments processed, pending @blake-regalia response.

blake-regalia:
> We are assuming now that they are valid `Terms`;
> Too complex and expensive otherwise.

@RubenVerborgh By your own logic, iff we are assuming they are valid `Terms` then we don't need to test for `null` or `undefined`! Where is the consistency in that?

RubenVerborgh:
Null pointer exceptions are a common source of errors, I expect several magnitudes more common than malformed Term exceptions.

blake-regalia:
> Null pointer exceptions are a common source of errors

Agreed - and I still think it would be nice to let devs know that the variable they passed to equals is actually `null` ;)  because who does that intentionally?

> I expect several magnitudes more common than malformed Term exceptions.

Well we are not just going for a majority of cases, it needs to be well-defined, consistent and fully-covered.

RubenVerborgh:
Alright, not disagreeing, but I don't have more bandwidth to follow up on this. Re-assigning to @bergos and @blake-regalia, feel free to conclude in the way you think is most appropriate.

elf-pavlik:
I think this relates to #104 (Nuance "plain objects (JSON) cannot be used) since @blake-regalia just used such 'plain objects' to make his case.

bergos:
We could add to the list of conditions:

> no error is thrown when evaluating the `.equals` calls

Then we would be on the safe side.

But also think of the impact. Maybe you want to throw an error on purpose. I don't have a good example now, but maybe access control could be a use case or a native implementation with a very low level error. You don't want to swallow errors you throw on purpose. 

From my point of view it's not about catching all errors (also not null pointer exceptions which could be thrown by a `Term` implementation), it's about allowing nothing (`null`/`undefined`) or a `Quad`. `String.prototype.include` also supports my point, as the following code throws an error:

`'a'.includes({toString: () => {throw Error()}})`

RubenVerborgh:
> > no error is thrown when evaluating the `.equals` calls
> 
> Then we would be on the safe side.

Too expensive to make such a guarantee.

vhf:
> From my point of view it's not about catching all errors […], it's about allowing nothing (`null`/`undefined`) or a `Quad`.

Agreed.

> iff we are assuming they are valid `Terms` then we don't need to test for `null` or `undefined`!

@blake-regalia I understand this concern about consistency.

On the one hand the `.equals` algorithm suggested in this PR doesn't explicitly state that `other` should be a term but the interface does:

```
interface Term {
[…]
  boolean equals(Term other);
};
```

On the other hand the `.equals` algorithm allows for `null` and `undefined` while the interface doesn't.

What would you all think of 

```diff
interface Term {
[…]
- boolean equals(Term other);
+ boolean equals(optional Term? other);
};
```

I think it would solve this particular issue: now it makes sense that the algo has to cover `null` and `undefined`, and that `other` can only be a `Term` when defined and not null.

elf-pavlik:
maybe @vhf you could make separate PR to the branch of this PR? Ruben mentioned having no more bandwidth left few comments above

blake-regalia:
@vhf I think that is a great suggestion!

blake-regalia:
@vhf In that case, I assume then it would be OK for implementations to `throw` if passed, for instance, a `string` or `number` which does not have a `.termType` property? I think then this algo would remove the criteria about checking that those properties exist at runtime, correct?

RubenVerborgh:
> a `string` or `number` which does not have a `.termType` property?

That won't happen unless you throw it yourself; all strings and numbers have such a property.

blake-regalia:
> That won't happen unless you throw it yourself;

To clarify, what I mean is: would throwing be allowed as opposed to returning false?

RubenVerborgh:
I’m opposed to mandating exceptions in this low-level spec for performance reasons. We could leave the behavior undefined, worst case.

blake-regalia:
Yes I agree and I am totally happy with leaving it undefined as long as we clarify that in the spec; i.e., this would mean that throwing or returning false are both allowed.

vhf:
I added a commit for the first part of the latest comments: making `.equals` parameter optional and nullable.

As for the other part I'm not sure to understand your suggestion @blake-regalia , let's start here:

> In that case, I assume then it would be OK for implementations to `throw`

Why this assumption? This PR so far doesn't let you throw when you implement `.equals`, and it specifies that `other` can be undefined or null, and how to handle the case when it's undefined, null, or a Term/Quad.

bergos:
@blake-regalia can you make a proposal what to change so the PR gets your approval?

RubenVerborgh:
Closing this concrete proposal until we reach an agreement here: https://github.com/rdfjs/representation-task-force/issues/151

RubenVerborgh:
Reopening based on resolution of #151.

elf-pavlik:
@RubenVerborgh with current approvals I think you can just go ahead and merge it.

blake-regalia:
I added @vhf for review as he had a good suggestion about the signatures.

RubenVerborgh:
No objection to that.

elf-pavlik:
@blake-regalia should we just apply patch @vhf posted in his comment above? https://github.com/rdfjs/data-model-spec/pull/142#issuecomment-457104605 and do manual merge?

blake-regalia:
Okay, made the commits -- let's wait for @vhf to review and then we can merge.

PR #153, opened 2019-02-07
remove Triple and .triple()

### LINKED ISSUE #124: Get rid of `Triple`? A case for the misleading class
Right now, `Triple` is an alias for `Quad` with its `.graph` property set to an instance of a `DefaultGraph`. The factory method `.quad()` can be called without supplying the optional final graph argument, in which case the `.graph` property defaults to the default graph. This makes the factory method `.triple()` redundant, but more importantly, misleading.

At first glance this seems trivial, but if I want to compare two triples from different graphs, then I would expect an instance of `Triple` to only compare subject, predicate and object. Without the distinction between triple and quad, implementations cannot extend the interface to allow for, say, extracting *an actual triple* from one graph and comparing it to a triple from another graph.

One solution would be getting rid of `Triple` entirely from the spec, including the factory method, and treating everything as a Quad. The other option would be to make the distinction that `Triple` is its own class that **extends** `Quad`. The latter of which is more supportive of implementations imo.

elf-pavlik:
closes #124


elf-pavlik:
7 days already and 3 approving reviews, would someone like to merge it?

Issue #150, opened 2019-01-28
Proposal: remove equals method


RubenVerborgh:
In https://github.com/rdfjs/representation-task-force/issues/104, there was a suggestion to drop the `.equals` method from `Term` and `Quad`. Let's express opinions with arguments here. Please be clear and brief.

Depends on https://github.com/rdfjs/representation-task-force/issues/151: the assumption for including an `equals` method at all is that we come to an agreement of what it would look like in https://github.com/rdfjs/representation-task-force/issues/151. Failure to reach such an agreement would likely also mean that `equals` has to be removed, regardless of the outcomes of this thread. So I would suggest to assume here that https://github.com/rdfjs/representation-task-force/issues/151 succeeds with one of the three possible outcomes in there. If you disagree with any of those outcomes, https://github.com/rdfjs/representation-task-force/issues/151 is probably be the right place.

Please vote with :+1: or :-1: on this message.

RubenVerborgh:
I vote 👎 because our goal with RDF/JS is to make an idiomatic JavaScript API. In such APIs, objects can have properties, methods, etc. The idiomatic OO way to check for equality, in absence of operator overloading, is a member method.

elf-pavlik:
I vote :+1: because I see different implementers having mutually exclusive expectations to how such member method should work. As I understand implementations can still provide such custom member method behaving as this implementation expects, at the same time libraries which aim to work only based on what the spec defines can still use a function of their choice which will behave in a way they expect.

I think we still haven't documented how different data factories supposed coexist (eg. https://github.com/rdfjs/representation-task-force/issues/79#issuecomment-457259704) and haven't followed up on converting instances created with one factory to instances of another factory https://github.com/rdfjs/representation-task-force/issues/137 Possibly having more clarity in that aspect will allow implementers to evaluate difference between having a 'blessed' `.equals()` member methods in the spec and having them as custom addition of specific implementation. Currently as I understand without 'blessed' `.equals()` implementation would need to:
* use function that takes two parameters and compares them
* ensure that all instances come from the factory which implements custom member method which matches their expectations

RubenVerborgh:
> I vote 👍 because I see different implementers having mutually exclusive expectations to how such member method should work ([chat comment](https://gitter.im/rdfjs/Representation-Task-Force?at=5c4b2d689bfa375aab5b3012)).

[...]

PR #158, opened 2019-02-15
DataFactory methods to create from existing term and quad | closes #137

### LINKED ISSUE #137: Add DataFactory method to create from existing term

elf-pavlik:
As discussed in #137

### DIFF
--- a/index.html
+++ b/index.html
@@ -372,12 +372,13 @@
       it MUST set <code>graph</code> to a <code>DefaultGraph</code>.
     </p>
     <p>
-      <dfn>fromTerm()</dfn> returns a new instance of specific <code>Term</code> subclass,
-      which passed to <code>original.equals()</code> returns <code>true</code>
+      <dfn>fromTerm()</dfn> returns a new instance of the specific <code>Term</code> subclass given by `original.termType`
+      (e.g., `NamedNode`, `BlankNode`, `Literal`, etc.),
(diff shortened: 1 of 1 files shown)


elf-pavlik:
:wave:

blake-regalia:
I'm pretty sure that with #142 we could simply say, such that `newObject.equals(original)` as this would cover all cases including *plain objects*, correct?

elf-pavlik:
Can we do it the other way and require that newly created instance will have `.equals()` method and passing `original` to it returns `true` ? Since it looks like spec will include definition of `.equals()` #150 I think that `fromTerm` and `fromQuad` definitions should simply take advantage of already eisting definition of `.equals()`

blake-regalia:
@elf-pavlik I think that is exactly what I am proposing... do we have a misunderstanding?

elf-pavlik:
Thanks for commit @blake-regalia :+1: 
No misunderstanding, just race condition on github comments :racehorse:

elf-pavlik:
@dmitrizagidulin I noticed you had some suggestions in https://github.com/rdfjs/data-model-spec/issues/81#issuecomment-239825876 would you like to review this PR?

PR #160, opened 2019-04-03
Clarify that returned `Terms` are novel

ericprud:
for s a quad store with one triple,
## Triples are novel
``` javascript
s.getQuads()[0].equals(s.getQuads()[0]) // true
s.getQuads()[0] == s.getQuads()[0] // false
```
## Terms are novel
``` javascript
s.getQuads()[0].subject.equals(s.getQuads()[0].subject) // true
s.getQuads()[0].subject == s.getQuads()[0].subject // false
```

### DIFF
--- a/index.html
+++ b/index.html
@@ -100,6 +100,9 @@
     <li>Interfaces do <em>not</em> specify how instances are stored in memory.</li>
     <li>Interfaces mandate specific pre-defined methods such as <code>.equals()</code>.</li>
     <li>
+       Interfaces are not expected to expose internal objects or return the same object from sucessive calls. Callers shoud use the `Term.equals()` function instead of `===` or `==` to test for equivalence. 
+    </li>
+    <li>
       Factory functions (e.g., <code>quad()</code>) or methods (e.g.,
       <code>store.createQuad()</code>) create instances.
     </li>


ericprud:
The problem with "not obligated" is that it lures folks into writing incompatible code, i.e. they develop on some system where Term identity is preserved and that code doesn't work with other RDF JS stores. It's fine if they do that for performance or simplicity, but it should be very clear to them when they do that they are stepping outside RDF JS and sacrificing interop.

RubenVerborgh:
I would rather add to the spec that equality checks with `==` or `===` are not supported; should always be `equals`.

elf-pavlik:
@RubenVerborgh I see this PR adds that implementations *should* use `equals()` do you have specific way of rephrasing it you would prefer?

RubenVerborgh:
Equality MUST only be checked with equals?

ericprud:
I guess I brought us into 2119-land with the use of "should". I think we're not saying that you're not *allowed* to use `==`, just that it's not RDF JS if you do. How about changing the last sentence:?
``` diff
- Callers should use the `Term.equals()` function instead of `===` or `==` to test for equivalence.
+ Equivalence is tested with the `Term.equals()` function instead of `===` or `==`.
```

RubenVerborgh:
👍

Issue #162, opened 2020-05-20
RDF* support


rubensworks:
[RDF*](https://blog.liu.se/olafhartig/2019/01/10/position-statement-rdf-star-and-sparql-star/) provides a way to annotate triples, and is quickly gaining traction by major SPARQL and RDF vendors+frameworks, such as [GraphDB](http://graphdb.ontotext.com/documentation/9.2/free/devhub/rdf-sparql-star.html) and [Jena](https://jena.apache.org/documentation/rdfstar/).

Since RDF/JS aims to be a common data model for RDF applications in JS,
we should consider adding RDF* support to the data model.
Deciding whether or not RDF* is the proper way forward for RDF should not be a concern here IMO.

To enable this, the only thing that would require a change AFAICS, is that the `Quad` interface should implement the `Term` interface (and a `"Quad"` `termType` should be added).

~~Before I open a PR for this, I first want to hear other people's thoughts on this topic.~~

-----

At the moment, two interface proposals seem to be receiving most attention:

* Composition (#163): `QuadTerm` is a `Term` containing a `Quad`.
* Extension (#165): `Quad` implements `Term`.

So far, none of them seem to jump out as a clear "best" approach.
In this comment, I will attempt to summarize the main discussion points (in no particular order), and how they relate to each approach.

|        													  | Composition | Extension |
| ----------------------------------------------------------- | ----------- | --------- |
| Fully backwards compatible with current RDF/JS spec version |             | ✅        |
| Follows RDF* idea that a quad is a term                     |             | ✅        |
| Dynamically check RDF* support on a factory 				  | ✅          |           |
| No optional fields 										  | ✅          |           |
| Alignment to Jena          								  | ✅          |           |
| Directly compatible with non-RDF* factories                 | ✅          |           |

Please let me know if there are any points that are missing from this table, or are missing.

RubenVerborgh:
Sounds good (also the proposed implementation), but we might want it as opt-in.
So not extend `Quad` itself, but rather… `QuadTerm` perhaps?

rubensworks:
> not extend Quad itself, but rather… QuadTerm perhaps?

Yep, that makes sense. Would also be better for backwards-compatibility.
This should align nicely with the TS typings.

bergos:
What about allowing `termType` to be also `undefined` for `Quad`? Would keep the spec small and simple and I don't see any drawback.

RubenVerborgh:
[...]

PR #163, opened 2020-05-20
Add experimental interfaces for RDF*

rubensworks:
This PR includes an initial draft for adding RDF* support as discussed in #162.

### DIFF
--- a/index.html
+++ b/index.html
@@ -51,6 +51,11 @@
         url: "https://ruben.verborgh.org/",
         company: "Ghent University – imec",
         companyURL: "http://idlab.ugent.be/"
+      }, {
+        name: "Ruben Taelman",
+        url: "https://www.rubensworks.net/",
+        company: "Ghent University – imec",
+        companyURL: "http://idlab.ugent.be/"
       }],
       bugTracker:             {
         open: "https://github.com/rdfjs/data-model-spec/issues?utf8=%E2%9C%93&q=is%3Aissue%20is%3Aopen%20",
@@ -144,8 +149,8 @@
     <p>
       <dfn>termType</dfn> contains a value that identifies the concrete interface of the term, since
       Term itself is not directly instantiated. Possible values include <code>"NamedNode"</code>,
-      <code>"BlankNode"</code>, <code>"Literal"</code>, <code>"Variable"</code> and
-      <code>"DefaultGraph"</code>.
+      <code>"BlankNode"</code>, <code>"Literal"</code>, <code>"Variable"</code>, <code>"DefaultGraph"</code>,
+      <code>"Quad"</code> and <code>undefined</code>.
     </p>
     <p>
       <dfn>value</dfn> is refined by each interface which extends Term.
@@ -364,6 +369,41 @@
       otherwise, it returns <code>false</code>.
     </p>
   </section>
+  
+  <section data-dfn-for="QuadTerm">
+    <h3><dfn>QuadTerm</dfn> interface</h3>
+    
+    <div class="warning" title="Experimental!">
+      <p>
+        The <code>QuadTerm</code> interface an experimental part of RDF* support, and is subject to change.
+      </p>
+    </div>
+
+    <pre class="idl">
+    interface QuadTerm : Quad {
+      attribute string termType;
+      attribute string value;
+      boolean equals(optional (Quad | Term)? other);
+    };
+    </pre>
+
+    <p>
+      An instance of <code>QuadTerm</code> is a <code>Quad</code> that is at the same time a <code>Term</code>.
+      This enables a <code>QuadTerm</code> to be used within a <code>Quad</code> or <code>QuadTerm</code>,
+      which enables annotation of statements as introduced by <a href="https://blog.liu.se/olafhartig/2019/01/10/position-statement-rdf-star-and-sparql-star/">RDF*</a>.
+    </p>
+    <p>
+      <dfn>termType</dfn> contains the constant <code>"Quad"</code>.
+    </p>
+    <p>
+      <dfn>value</dfn> contains an empty string as constant value.
+    </p>
+    <p>
+      <dfn>equals()</dfn> returns <code>true</code> if
+      all <a>Quad.equals</a> and all general <a>Term.equals</a> conditions hold;
+      otherwise, it returns <code>false</code>.
+    </p>
+  </section>
 
   <section data-dfn-for="DataFactory">
     <h3><dfn>DataFactory</dfn> interface</h3>
@@ -418,6 +458,39 @@
     <p>
       <dfn>fromQuad()</dfn> returns a new instance of <code>Quad</code>,
       such that <code>newObject.equals(original)</code> returns <code>true</code>.
+    </p>
+  </section>
+
+  <section data-dfn-for="DataFactory">
+    <h3><dfn>DataFactoryStar</dfn> interface</h3>
+    
+    <div class="warning" title="Experimental!">
+      <p>
+        The <code>DataFactoryStar</code> interface an experimental part of RDF* support, and is subject to change.
+      </p>
+    </div>
+
+    <pre class="idl">
+    interface DataFactoryStar : DataFactory {
+      QuadTerm quad(Term subject, Term predicate, Term object, optional Term? graph);
+      QuadTerm fromQuad(Quad original);
+    };
+    </pre>
+
+    <p>
+      The <code>DataFactoryStar</code> interface extends the <code>DataFactory</code> interface,
+      and overrides the <code>quad</code> and <code>fromQuad</code> methods.
+    </p>
+    <p>
+      <dfn>quad()</dfn>returns a new instance of <code>QuadTerm</code>.
+      If <code>graph</code> is <code>undefined</code> or <code>null</code>
+      it MUST set <code>graph</code> to a <code>DefaultGraph</code>.
+    </p>
+    <p>
+      <dfn>fromQuad()</dfn> returns a new instance of <code>QuadTerm</code>.
+      such that <code>original.equals(newObject)</code> returns <code>true</code>.
+      <code>newObject.equals(original)</code> will only return true
+      if <code>original</code> was already an instance of <code>QuadTerm</code>.
     </p>
   </section>
 </section>
--- a/index.html
+++ b/index.html
@@ -138,7 +138,7 @@
     <pre class="idl">
     interface Term {
       attribute string termType;
-      attribute string value;
+      attribute (string | Quad) value;
       boolean equals(optional Term? other);
     };
     </pre>
@@ -150,7 +150,7 @@
       <dfn>termType</dfn> contains a value that identifies the concrete interface of the term, since
       Term itself is not directly instantiated. Possible values include <code>"NamedNode"</code>,
       <code>"BlankNode"</code>, <code>"Literal"</code>, <code>"Variable"</code>, <code>"DefaultGraph"</code>,
-      <code>"Quad"</code> and <code>undefined</code>.
+      <code>"Quad"</code>.
     </p>
     <p>
       <dfn>value</dfn> is refined by each interface which extends Term.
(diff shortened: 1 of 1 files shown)


bergos:
The goal of my proposal to allow `undefined` in the `termType` was to use only the `Quad` interface and not require a `QuadTerm` interface. RDF is simple and RDF* is almost as simple. If possible I would like to keep the number of interfaces low. My thoughts from two perspectives:

From developers that use the interfaces:
- Easy to understand because of the small number of interfaces that match almost exactly the RDF model 👍 
- RDF* can be found via search or maybe even a separate section 👍 
- No need to go into details and understand why there is a `QuadTerm` interface besides the `Quad` interface 👍 
- RDF* and the fact that `Quad` implements `Term` is a little bit hidden 👎 

From developers that implement the interface:
- They are experts, they know what to look for and they have to understand all details 👍 
- I guess they explicitly decide to support RDF* or decide against it, so no need to have two separate factories 👍 
- Many use cases support only a subset of the valid values of `termType` and therefore must validate the value already. That happens outside the data model implementation. e.g. `Variable` shows up at a serializer 👍 

I only see one drawback, but a separate section about RDF* could cover that point.

rubensworks:
> use only the Quad interface and not require a QuadTerm interface

I don't immediately see any downsides of that approach.
I'll wait for @RubenVerborgh's review before modifying this PR.

> I only see one drawback, but a separate section about RDF* could cover that point.

A separate section dedicated to RDF* is a good idea, will add.

index.html
```diff
@@ -144,8 +149,8 @@ <h3><dfn>Term</dfn> interface</h3>
     <p>
       <dfn>termType</dfn> contains a value that identifies the concrete interface of the term, since
       Term itself is not directly instantiated. Possible values include <code>"NamedNode"</code>,
-      <code>"BlankNode"</code>, <code>"Literal"</code>, <code>"Variable"</code> and
-      <code>"DefaultGraph"</code>.
+      <code>"BlankNode"</code>, <code>"Literal"</code>, <code>"Variable"</code>, <code>"DefaultGraph"</code>,
+      <code>"Quad"</code> and <code>undefined</code>.
```
RubenVerborgh, line 153: I'm not a big fan of the `undefined` here.
By doing it this way, we propagate an RDF* concern to everywhere.

I would just leave it to the string values, and then on some interface declare it as optional.

index.html
```diff
@@ -364,6 +369,41 @@ <h3><dfn>Quad</dfn> interface</h3>
...
+    <h3><dfn>QuadTerm</dfn> interface</h3>
+    
+    <div class="warning" title="Experimental!">
+      <p>
+        The <code>QuadTerm</code> interface an experimental part of RDF* support, and is subject to change.
+      </p>
+    </div>
+
+    <pre class="idl">
+    interface QuadTerm : Quad {
```
RubenVerborgh, line 383: I had expected for it to be (an alias for) both a `Quad` and a `Term` actually.

index.html
```diff
@@ -364,6 +369,41 @@ <h3><dfn>Quad</dfn> interface</h3>
...
+      attribute string termType;
+      attribute string value;
+      boolean equals(optional (Quad | Term)? other);
+    };
+    </pre>
+
+    <p>
+      An instance of <code>QuadTerm</code> is a <code>Quad</code> that is at the same time a <code>Term</code>.
```
RubenVerborgh, line 391: See 🙂

index.html
```diff
@@ -420,6 +460,39 @@ <h3><dfn>DataFactory</dfn> interface</h3>
...
+    
+    <div class="warning" title="Experimental!">
+      <p>
+        The <code>DataFactoryStar</code> interface an experimental part of RDF* support, and is subject to change.
+      </p>
+    </div>
+
+    <pre class="idl">
+    interface DataFactoryStar : DataFactory {
+      QuadTerm quad(Term subject, Term predicate, Term object, optional Term? graph);
```
RubenVerborgh, line 475: Do we really want to overwrite? I suppose we do, just wanted to hear arguments.

RubenVerborgh:
I reviewed, but my comments seem orthogonal to @bergos'. I'm also not sure I fully follow; could we write out some of the differences for better comparison?

retog:
I would prefer composition rather than inheritance, i.e. having `QuadTerm` a Term with the quad as its value and "quad" as termType. This approach would also not require the ugly `undefined` and make the design generally less complex.

rubensworks:
Thank you all for your comments.

I played around with different possibilities in TypeScript, and it felt like a composition-based led to the most straightforward solution (thanks for the reminder @retog).

I have pushed a reworked version of the spec, and here you can find the corresponding typings (which can be pushed once if proposal PR is accepted): https://github.com/rubensworks/DefinitelyTyped/commit/955b974ce991440d651ad8bf3ffd7534705f2e15
Please let me know what you think.

bergos:
@rubensworks I think the interfaces don't match with the RDF* model. The `Quad` is the `Term` and not the value of a `QuadTerm`. We had a similar discussion about the `NamedNode` interface. A `NamedNode` is a `Term` that has an `IRI` as the value. The discussion can be found in #50. I think also here we should stay consistent with the rest of the interfaces and the RDF* model.

As @RubenVerborgh proposed I created two PRs for better comparison. #164 defines a new interface, #165 extends the `Quad` interface.

I still favor the one with the extended `Quad` interface. It's small, simpler, and easier to understand for developers that use the interfaces. As I mentioned earlier we can have two different perspectives on the specification: Developers that use the interface or developers that implement the interface. As I don't expect there will be a separate primer, it would be good to have a low entry barrier. #165 would only require to know the RDF model, but not the RDF* model to understand the basic concept of all interfaces.

rubensworks:
Thanks for the additional PRs @bergos.

> The Quad is the Term and not the value of a QuadTerm

I may be misunderstanding #50, but this feels like a different problem. I see no issue in making `Quad` the value of a `QuadTerm`. This is similar as to how the datatype of a `Literal` is defined.


The additional datafactory method in this PR can be seen as an advantage IMO, as it makes RDF* a clear opt-in.
In any case, I don't have a strong preference, I would also be fine with #165. (#164 seems a bit confusing).

rubensworks:
@rdfjs/data-model-spec @retog Any other comments on this PR and #164+#165?
As said before, I have no strong preference.

I hope we can decide on an approach soon, as we have some students starting this month, and I hope to let them work on some RDF* tooling, which would benefit all of us.

tpluscode:
I think I be in favour of #165. 

```
<<:bob foaf:age 23>> ex:certainty 0.9
```

This appears to closely match the Quad being a Term standpoint

```
```

bergos:
@rubensworks 

> I may be misunderstanding #50, but this feels like a different problem. I see no issue in making Quad the value of a QuadTerm. This is similar as to how the datatype of a Literal is defined.

I mentioned it in the context of keeping the structures of the RDF model also in the RDF/JS structures. The RDF* model makes the `Quad` a `Term` and doesn't introduce a new kind of `Term` with a `Quad` as the value. Like the `Literal`, a `Quad` has properties with known names.

> The additional datafactory method in this PR can be seen as an advantage IMO, as it makes RDF* a clear opt-in.

Libraries are always free to export different factories. On purpose, the spec doesn't cover that part.

RubenVerborgh:
I think this is the way to go; it also aligns with Jena: https://jena.apache.org/documentation/rdfstar/

Proposal: let's start with this option, marked as experimental, and get implementer feedback.

tpluscode:
Let's compare from proposed usage perspective. Code to create a quad like above

```turtle
<<:bob foaf:age 23>> ex:certainty 0.9 .
```

With `this` approach:

```js
import { namedNode, literal, quad, quadTerm } from '@rdfjs/data-model'

const bobAgeQuad = quad(
  namedNode(bob),
  foaf.age,
  literal('23', xsd.int)
)

const quad = quad(
  quadTerm(bobAgeQuad),
  ex.certainty,
  literal('0.9', xsd.decimal)
)

console.log(`
  <<
    ${quad.subject.value.subject} ${quad.subject.value.predicate} ${quad.subject.value.object}
  >> ${quad.predicate} ${quad.object}`
)
```

Compared with changes proposed in #165 

```diff
-import { namedNode, literal, quad, quadTerm } from '@rdfjs/data-model'
+import { namedNode, literal, quad } from '@rdfjs/data-model'

const bobAgeQuad = quad(
  namedNode(bob),
  foaf.age,
  literal('23', xsd.int)
)

quad(
- quadTerm(bobAgeQuad),
+ bobAgeQuad,
  ex.certainty,
  literal('0.9', xsd.decimal)
)

console.log(`
  <<
-  ${quad.subject.value.subject} ${quad.subject.value.predicate} ${quad.subject.value.object}
+  ${quad.subject.subject} ${quad.subject.predicate} ${quad.subject.object}
  >> ${quad.predicate} ${quad.object}`
)
```

IMO the simplicity speaks in favour of #165

RubenVerborgh:
Simplicity can be added in layers above; RDF/JS is about low-level library interop, where we need the most solid model.

However, can we have best of both worlds?
Why don't we make a quad _always_ a term? I don't like the optional fields in #165, they make things messy as per my first sentence here.

tpluscode:
> RDF/JS is about low-level library interop

I don't think this is true. RDF(/JS) is the lingua franca of all tools, both high and low level. I find myself using the data factory every time in different places of the applications I implement. It is important to me if it will require a new `quadTerm` import and additional `.value` to access the RDF*-reified quads

> Why don't we make a quad always a term?

From #165 

```diff
-    interface Quad {
+    interface Quad : Term {
```

"always term", no?

RubenVerborgh:
> RDF(/JS) is the lingua franca of all tools

Lingua franca, which means that any abstraction built on top of RDF/JS will work with any of the libraries.

Hence, the goal is to find a solid and minimal interoperable set, not necessarily the simplest interface. That's up to other libraries.

tpluscode:
> Lingua franca, which means that any abstraction built on top of RDF/JS will work with any of the libraries.
> (...)
> That's up to other libraries. 

I may have not been precise enough.

It is not only high level tool that you use the RDF/JS APIs. The RDF model easily surfaces to all layers regardless of the libraries used. What we do here I would compare to attempting to change JSON itself in non-RDF land. I opt for least impact.  #165 proposes least invasive changes

> the goal is to find a solid and minimal interoperable set

Could you elaborate what you don't like about #165? (other than the optional `termType` and `value` which I agree with you and commented)

Here's my take, expanding on what @bergos laconically mentions above. In RDF* a quad can **be** the subject of quad. Simple as that, the quad itself **is a term**. This is most accurately represented in that PR. Not a term, whose value is a quad.

RubenVerborgh:
> other than the optional `termType` and `value` which I agree with you and commented)

That seems to be integral to the design; so yes, that.

bergos:
> I think this is the way to go; it also aligns with Jena: https://jena.apache.org/documentation/rdfstar/

That's just the SPARQL result, it's way more complicated in the Jena API and we should not try to align to it...

> Why don't we make a quad always a term?

I don't have a problem with it, I would even implement it that way. I had the impression that some people want to be able to explicitly say a Quad is not a Term, if they don't use RDF*, but I don't see any drawbacks to always make a Quad a Term. The only benefit would be to be backward compatible, but then there are libraries implementing version 1.0 of the spec and others 1.1. Should not be a problem.

rubensworks:
To recap, this PR initially looked like #165, but I changed this to be based on composition (following @retog's suggestion) after playing around with it in TS. The composition approach felt more robust, as it did not have the problem of optional fields on terms.

With this composition approach, you for example do not have any problems if you want to create RDF* statements based on quads that were obtained from some source that is out of your control, and makes use of an old datafactory (not supporting RDF*):

```typescript
const quads: RDF.Quad[] = obtainQuadsFromSomewhere();
const quadTerm = myDataFactory.createQuadTerm(quads[1]);
myDataFactory.quad(quadTerm, ...);
```

With the approach from #165, this would not be possible without re-creating the quad.

> That's just the SPARQL result, it's way more complicated in the Jena API and we should not try to align to it...

Jena's Model API  for RDF* on the bottom of [that page](https://jena.apache.org/documentation/rdfstar/) doesn't look that complicated to me:

```
    Statement Resource.getStatement()
    Resource Model.createResource(Statement)
    Resource ResourceFactory.createStatement
```

Alignment to this composition approach definitely makes sense to me.

tpluscode:
> With this composition approach, you for example do not have any problems if you want to create RDF* statements based on quads that were obtained from some source that is out of your control

Why is this a problem exactly? The quads are immutable so it doesn't really make much of a difference.

And from pure memory consumption/allocation it is also hardly relevant. Either way the RDF* factory will have to create new objects for every reified quad. For every quad subject It will be either a `QuadTerm` or re-packaged `Quad`. 

Unless we count down to individual allocated bytes, I'd say that the practical difference in negligible

bergos:
> The composition approach felt more robust, as it did not have the problem of optional fields on terms.

As I mentioned before, we can also make them mandatory. `termType` is fixed `Quad` and `value` fixed an empty string, just like we defined it for the `DefaultGraph`.

> With this composition approach, you for example do not have any problems if you want to create RDF* statements based on quads that were obtained from some source that is out of your control, and makes use of an old datafactory (not supporting RDF*):

That's the reason why we pass around factories.

> Jena's Model API for RDF* on the bottom of that page doesn't look that complicated to me:

I went down the rabbit hole and it looks like everything is stored in a property called [`label`](https://github.com/apache/jena/blob/master/jena-core/src/main/java/org/apache/jena/graph/Node.java#L42). It is way more complicated and the factory concept is different from RDF/JS. Why should we align with a typed language if we are more free to align it to the actual model? I see that contrary to the initial idea of having a idiomatic API for JavaScript.

index.html
```diff
@@ -133,7 +138,7 @@ <h3><dfn>Term</dfn> interface</h3>
     <pre class="idl">
     interface Term {
       attribute string termType;
-      attribute string value;
+      attribute (string | Quad) value;
```
RubenVerborgh, line 141: We have just begun implementation in https://github.com/rdfjs/N3.js/pull/207, where I have noticed that this bothers me. We are violating object homomorphism, which might lead to slowdowns all around. The predictability of the `Term` structure enables high performance; it it can be either of two fundamentally different objects, we'll see issues.

tpluscode: A pull request 🤔. So you commit to this proposal even though it has not been decided which design to choose?

Given your comment, does that mean that maybe you would be in favour of #165 after all?

RubenVerborgh: > So you commit to this proposal

No, we're experimenting with the various options to provide feedback. And good that we are, because that's what led to my comment above 🙂

rubensworks:
Closing in favor of #165.

PR #164, opened 2020-06-19
Added RDF* support with new QuadTerm interface

bergos:
This PR adds RDF* support with a new `QuadTerm` interface. The `Term` was changed to a mixin interface to allow the `QuadTerm` to inherit from `Quad` and also include the `Term` interface. All `Quad` types in the `DataFactory` have been replaced by a union type of `Quad` and `QuadTerm`.

### DIFF
--- a/index.html
+++ b/index.html
@@ -131,7 +131,7 @@
     <h3><dfn>Term</dfn> interface</h3>
 
     <pre class="idl">
-    interface Term {
+    interface mixin Term {
       attribute string termType;
       attribute string value;
       boolean equals(optional Term? other);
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -365,6 +375,67 @@ <h3><dfn>Quad</dfn> interface</h3>
     </p>
   </section>
 
+  <section data-dfn-for="QuadTerm">
+    <h3><dfn>QuadTerm</dfn> interface</h3>
+
+    <pre class="idl">
+    interface QuadTerm : Quad {
+      attribute string termType;
+      attribute string? value;
```
RubenVerborgh, line 384: Can we make this non-optional? We need predictable objects for high performance.

bergos: I changed it in #165. The approach in #165 got more positive feedback than this one. If you have good arguments to continue with this PR, I can change it, but first, it would be good to know if it's worth the effort.

RubenVerborgh: Just trying to understand/compare the different options (following https://github.com/rdfjs/data-model-spec/pull/163/files#r455358833).

So I will review this one negatively for this issue; that might help us with selecting. (i.e., I could still champion this one if the optional issue is resolved)

PR #165, opened 2020-06-19
Added RDF* support by extending Quad from Term

bergos:
This PR adds RDF* support by extending Quad from Term.

Please leave feedback till 2020-07-21 if you request any changes.
Please vote till 2020-07-28 👍  if this PR should be merged or 👎  if you prefer a different approach.

Below is a comparison of the [composition PR](https://github.com/rdfjs/data-model-spec/pull/163) and this PR.

|                                                                 | Composition | Extension |
| --------------------------------------------------------------- | ----------- | --------- |
| Fully backwards compatible with current RDF/JS spec version     |             | ✅        |
| Follows RDF* idea that a quad is a term                         |             | ✅        |
| No optional fields                                              | ✅          | ✅         |
| Dynamically check RDF* support on a factory [1]                 | ✅          | ✅         |

The following points have been excluded as they are contrary to the concept of the RDF/JS specifications:

- Alignment to Jena [2]
- Directly compatible with non-RDF* factories [3]

[1] Support for RDF* can be detected with `quad.termType === 'Quad'` on a `Quad` object, but see also [3].
[2] RDF/JS specifications are defined to be idiomatic for JavaScript.
[3] The factory interface exists to have the possibility to pass a specific factory based on the application requirements to other libraries without the need to recreated the objects. The specification explicitly mentions implementation specific properties, but of course, this also applies to features defined in minor versions of the specification.

### DIFF
--- a/index.html
+++ b/index.html
@@ -144,8 +144,8 @@
     <p>
       <dfn>termType</dfn> contains a value that identifies the concrete interface of the term, since
       Term itself is not directly instantiated. Possible values include <code>"NamedNode"</code>,
-      <code>"BlankNode"</code>, <code>"Literal"</code>, <code>"Variable"</code> and
-      <code>"DefaultGraph"</code>.
+      <code>"BlankNode"</code>, <code>"Literal"</code>, <code>"Variable"</code>,
+      <code>"DefaultGraph"</code> and <code>"Quad"</code>.
     </p>
     <p>
       <dfn>value</dfn> is refined by each interface which extends Term.
@@ -319,7 +319,9 @@
     <h3><dfn>Quad</dfn> interface</h3>
 
     <pre class="idl">
-    interface Quad {
+    interface Quad : Term {
+      attribute string? termType;
+      attribute string? value;
       attribute Term subject;
       attribute Term predicate;
       attribute Term object;
@@ -329,8 +331,15 @@
     </pre>
 
     <p>
-      <dfn>subject</dfn> the subject, which is a <code>NamedNode</code>, <code>BlankNode</code> or
-      <code>Variable</code>.
+      <dfn>termType</dfn> contains the constant <code>"Quad"</code> or <code>undefined</code> if the
+      implementation does not support RDF*.
+    </p>
+    <p>
+      <dfn>value</dfn> contains the constant <code>undefined</code>.
+    </p>
+    <p>
+      <dfn>subject</dfn> the subject, which is a <code>NamedNode</code>, <code>BlankNode</code>,
+      <code>Variable</code> or <code>Quad</code>.
     </p>
     <p>
       <dfn>predicate</dfn> the predicate, which is a <code>NamedNode</code> or
--- a/index.html
+++ b/index.html
@@ -320,8 +320,8 @@
 
     <pre class="idl">
     interface Quad : Term {
-      attribute string? termType;
-      attribute string? value;
+      attribute string termType;
+      attribute string value;
       attribute Term subject;
       attribute Term predicate;
       attribute Term object;
@@ -331,11 +331,10 @@
     </pre>
 
     <p>
-      <dfn>termType</dfn> contains the constant <code>"Quad"</code> or <code>undefined</code> if the
-      implementation does not support RDF*.
+      <dfn>termType</dfn> contains the constant <code>"Quad"</code>.
     </p>
     <p>
-      <dfn>value</dfn> contains the constant <code>undefined</code>.
+      <dfn>value</dfn> contains an empty string as constant value.
     </p>
     <p>
       <dfn>subject</dfn> the subject, which is a <code>NamedNode</code>, <code>BlankNode</code>,


index.html
```diff
@@ -319,7 +319,9 @@ <h3><dfn>DefaultGraph</dfn> interface</h3>
     <h3><dfn>Quad</dfn> interface</h3>
 
     <pre class="idl">
-    interface Quad {
+    interface Quad : Term {
+      attribute string? termType;
+      attribute string? value;
```
RubenVerborgh, line 324: But what would the value be?

tpluscode: I would not make these fields optional.

`termType` can easily be defined as `Quad` and it is not a breaking change to an implementation which does not support RDF*. They will simply ignore it

`value`, I'm not sure. Just make it a constant `undefined` or empty string?

rubensworks: I would say making these fields required _is_ a breaking change. Existing datafactories will be _required_ to expose them, and existing consumers of datafactories can break if old datafactories are used.

Having a `value` field here with a constant string or `undefined` feels too much like a hack to me, so I would not be in favour of that.

tpluscode: I think you're looking at the breaking change backwards.

Yes, a factory which upgrades to RDF* version will need to adjust (either adding new fields to `Quad` or adding `DataFactory#quadTerm` method.

Those are not important however. What is important is whether a library which depends on the current model spec would break once it is enhanced to support RDF*.

The answer is no, AFAICT.

rubensworks: Good point, I was looking at it backwards.

Only the `value` issue remains then.

tpluscode: Ha, and I was just about to reply similarly, having read your [other comment](https://github.com/rdfjs/data-model-spec/pull/163#issuecomment-656543500)

Trying to use "old quads" in RDF* factory indeed is not going to be seamless 🤔

index.html
```diff
@@ -329,8 +331,15 @@ <h3><dfn>Quad</dfn> interface</h3>
     </pre>
 
     <p>
-      <dfn>subject</dfn> the subject, which is a <code>NamedNode</code>, <code>BlankNode</code> or
-      <code>Variable</code>.
+      <dfn>termType</dfn> contains the constant <code>"Quad"</code> or <code>undefined</code> if the
+      implementation does not support RDF*.
+    </p>
+    <p>
+      <dfn>value</dfn> contains the constant <code>undefined</code>.
```
tpluscode, line 338: re @rubensworks 

> Having a value field here with a constant string or undefined feels too much like a hack to me, so I would not be in favour of that.

How about this?

```suggestion
      <dfn>value</dfn> contains a concatenation of the <code>value</code> properties of quad's subject, predicate, object and graph, separated by a single space character.
```

rubensworks: Not sure what the use of this value would be. I think it would just take up memory needlessly.

tpluscode: What is the use of value anyway? Would not have to be allocated up front. Make is a lazy getter?

rubensworks: Indeed, it doesn't have a value, so it shouldn't exist here IMO. But the `Term` contract currently requires it. Making it optional in `Term` would also be possible, but that would lead to other problems, as consumers will never be certain that this field exist on a `Term`.

tpluscode: > Indeed, it doesn't have a value

Clearly the quad **is** the value, no? As much as URI is the value of NamedNode

It was not what I asked. I meant to ask how is `Term#value` used. It has no use for equality and it is not representative of the actual value either (the way it ignores language/datatype of literals). Maybe you attribute too much importance to it right now?

> Making it optional in Term would also be possible, but that would lead to other problems

Well, yes, that would be a breaking change

rubensworks: > Clearly the quad is the value, no?

Exactly. This would be a motivation for the composition approach in #163 🙂

tpluscode: Fair enough

tpluscode: Ah, but no

```diff
interface Term {
- attribute string value;
+ attribute (string | Quad) value;
}
```

Changing the return type of `Term#value` is even worse. This actually is a breaking change because it is a hazard for code relying on the current interface where they will expect the `value` to be a string.

rubensworks: Yep, that would be a breaking change. There's no way around that in #163 AFAICS.

bergos:
Based on the feedback, this PR has been updated.
Making `termType` and `value` optional was the main concern.
`termType` is now defined to have the fixed value `Quad` and `value` must have the fixed value empty string.
The description has been updated which includes an updated version of the comparison of composition vs extension.

Based on the upvotes in the discussion it looks like many people are in favor of the approach covered in this PR.
I think it makes sense now to vote if this PR should be merged.

rubensworks:
Thanks for the changes @bergos, looks much better now!

In any case, I don't have a strong opinion which option we go for. As after playing around with both approaches in TS, the composition approach felt more natural and robust, my vote will go for that one.

Since we don't have an agreement on [2] and [3], I would appreciate it if you could add these back into the table of comparison. At the moment, it feels like a very biased comparison, compared to what I tried to do in #162.

Should this PR get merged, I would recommend adding an experimental note block and reference to the RDF* webpage like in #163.

RubenVerborgh:
This proposal has been implemented in N3.js by @simonvbrae as https://github.com/rdfjs/N3.js/pull/210. (Parser following in https://github.com/rdfjs/N3.js/pull/212.)

RubenVerborgh:
Released data model with parser and serializer support in `n3@1.6.0`.

Note that the RDF/JS community is of course still free to continue the debate; we just have implemented the proposal that we found most promising. Happy to adjust. Feedback welcome.

larsrh:
Continuing the conversation from rdfjs/N3.js#210:

It may be a good idea to introduce some form of property (or something else?) to be able to distinguish quad from non-quad terms. This is particularly useful in a typed setting. (Currently, the rdfjs types determine that something is a quad by the absence of the `termType` property.)

rubensworks:
@larsrh The introduction of `termType` on `Quad` feels like a clean way to check whether or not something is a quad, so I would say this PR already solves this problem.

@bergos So far, no new concerns have been reported to this PR, so I think this PR can be merged. Successful implementations in both N3.js and SPARQL.js confirm that this is a good way to go forward. 
However, as I said before, I would also add an experimental note block and reference to the RDF* webpage like in #163.

bergos:
Thanks for the feedback and voting. Based on the votes I merged this PR.

We may/should add some explanation for RDF* to the spec. Please leave some feedback in the issue #165. I will make a separate PR for the details which should be merged before we release a new version of the spec.

Issue #172, opened 2023-05-25
[RDF1.2] Add text direction to literals


rubensworks:
There's a high chance of [text directions](https://github.com/w3c/rdf-star-wg/wiki/Text-Direction-Proposal) to be added in RDF 1.2, so that string literals can indicate their direction (relevant I18N).

This issue is a place do discuss how we want to cope with this at RDF/JS data model-level.

In order to cope with this, we should add a new field to literals (e.g. `direction`), which may either be `'ltr'` or `'rtl'`.
The datafactory will also have to be modified to cope with this.
If we make `direction` optional, this should be fully backwards-compatible.

rubensworks:
For reference, this has now been merged in the rdf-concepts spec (WD): https://github.com/w3c/rdf-concepts/commit/f299fd0500628ead4b5f2fad23c5eab16d04d1c0

PR #175, opened 2023-10-25
Add optional direction for literals

rubensworks:
As mentioned in #172, RDF 1.2 is scheduled to introduce the [base direction concept](https://w3c.github.io/rdf-concepts/spec/#dfn-base-direction) for literals.

This PR proposes a backwards-compatible addition to the data model spec to handle this addition in RDF.

The literal interface change is quite straightforward, I don't expect much discussion on this.

The data factory on the other hand might deserve some discussion. I went for extending the types of the second optional argument of the `literal()` method. An alternative might have been to add a third optional argument, but I didn't go for that in this PR due to the possibility of defining a direction in combination with a datatype (which is illegal according to the RDF 1.2 spec).

### DIFF
--- a/index.html
+++ b/index.html
@@ -225,6 +225,7 @@
       attribute DOMString termType;
       attribute DOMString value;
       attribute DOMString language;
+      attribute DOMString? direction;
       attribute NamedNode datatype;
       boolean equals(optional Term? other);
     };
@@ -242,10 +243,16 @@
       <code>"en"</code>, <code>"en-gb"</code>) or an empty string if the literal has no language.
     </p>
     <p>
+      <dfn>direction</dfn> is defined if the string is a directional language-tagged string.
+      In this case, the <code>direction</code> MUST be either be <code>"ltr"</code> or <code>"rtl"</code>.
+    </p>
+    <p>
       <dfn>datatype</dfn> a <code>NamedNode</code> whose IRI represents the datatype of the literal.
     </p>
     <p>
-      If the literal has a language, its datatype has the IRI
+      If the literal has a language and a direction, its datatype has the IRI
+      <code>"http://www.w3.org/1999/02/22-rdf-syntax-ns#dirLangString"</code>.
+      If the literal has a language without direction, its datatype has the IRI
       <code>"http://www.w3.org/1999/02/22-rdf-syntax-ns#langString"</code>. Otherwise, if no
       datatype is explicitly specified, the datatype has the IRI
       <code>"http://www.w3.org/2001/XMLSchema#string"</code>.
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -404,10 +428,13 @@ <h3><dfn>DataFactory</dfn> interface</h3>
...
-      <dfn>literal()</dfn> returns a new instance of <code>Literal</code>. If
-      <code>languageOrDatatype</code> is a <code>NamedNode</code>, then it is used for the value of
-      <code>datatype</code>. Otherwise <code>languageOrDatatype</code> is used for the value of
-      <code>language</code>.
+      <dfn>literal()</dfn> returns a new instance of <code>Literal</code>.
+      If <code>languageOrDatatype</code> is a <code>NamedNode</code>,
+      then it is used for the value of <code>datatype</code>.
+      If <code>languageOrDatatype</code> is a <code>DirectionalLanguage</code>,
+      then its <code>language</code> and <code>direction</code> attributes
+      are respectively used for the literal's <code>language</code> and <code>direction</code>.
```
bergos, line 436:
```suggestion
      are respectively used for the literal's <code>language</code> and <code>direction</code>, where <code>direction</code> is optional or can be falsy.
```

Just to make it clearer, that direction is optional or can be falsy.

index.html
```diff
@@ -241,11 +252,17 @@ <h3><dfn>Literal</dfn> interface</h3>
       <dfn>language</dfn> the language as lowercase [[BCP47]] string (examples:
       <code>"en"</code>, <code>"en-gb"</code>) or an empty string if the literal has no language.
     </p>
+    <p>
+      <dfn>direction</dfn> is defined if the string is a directional language-tagged string.
```
bergos, line 256:
```suggestion
      <dfn>direction</dfn> is not falsy if the string is a directional language-tagged string.
```

index.html
```diff
@@ -254,7 +271,8 @@ <h3><dfn>Literal</dfn> interface</h3>
       <dfn>equals()</dfn> returns <code>true</code> if
       all general <a>Term.equals</a> conditions hold,
       <code>term.value</code> is the same string as <code>other.value</code>,
-      <code>term.language</code> is the same string as <code>other.language</code>, and
+      <code>term.language</code> is the same string as <code>other.language</code>,
+      <code>term.direction</code> is the same string as <code>other.direction</code> or are both undefined, and
```
bergos, line 275:
```suggestion
      <code>term.direction</code> is the same string as <code>other.direction</code> or are both falsy, and
```

bergos (changes requested):
Can we define that `direction` must be falsy instead of `undefined` for other kinds of literals? That would allow aligning `direction` to `language` with an empty string and still be backward compatible.

bergos (approved):
LGTM

rubensworks:
Will merge this as soon as https://github.com/rdfjs/types/pull/38 is merged (on hold now).

bergos:
Maybe a more detailed description is required if it's unclear why an empty string is allowed:

We had a longer discussion about whether the language property should be null or empty string for non-language literals and decided to go with the empty string, so it's possible to call .toUpperCase() or slice it, even if no language was set.

Aligning the direction property with the language property without having a breaking change required to allow undefined. Otherwise, I would have proposed to skip allowing undefined.

Now we have the optional chaining operator in JavaScript, and that would be a good reason to use undefined or null instead of an empty string. But as long as we don't have a strong need for a breaking change, I would try to keep it consistent. That also means a factory that supports direction should an empty string instead of null or undefined.

@rubensworks I can create a proposal if you agree to update the description.

rubensworks:
@bergos Sounds good to me!

PR #176, opened 2024-08-27
Replace _object with object

nichtich:
This seems to be a typo.

### DIFF
--- a/index.html
+++ b/index.html
@@ -326,7 +326,7 @@
       attribute DOMString value;
       attribute Term subject;
       attribute Term predicate;
-      attribute Term _object;
+      attribute Term object;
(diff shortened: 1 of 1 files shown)


rubensworks:
I believe this is intentional due to ReSpec limitations. But @bergos will probably know for certain.

bergos:
@rubensworks almost correct. It's defined in the [WebIDL specification](https://webidl.spec.whatwg.org/#idl-names):

> Note: A leading U+005F (_) is used to escape an identifier from looking like a reserved word so that, for example, an interface named "interface" can be defined. The leading U+005F (_) is dropped to unescape the identifier.

PR #182, opened 2026-02-26
Add immutability requirement for Term and Quad

### LINKED ISSUE #81: immutable objects
Are objects created by the `DataFactory` immutable? How should we write that in the spec? I think we said we don't define the properties read-only to avoid forcing the usage of `Object.defineProperty` in the implementations.

smessie:
I recently experienced weird and unexpected behavior when I was updating the values of terms in my project.
In the end, it appears that it was decided [in the meeting of 2016/10/13](https://github.com/rdfjs/data-model-spec/wiki/Meeting-2016-10-13) that terms and quads are actually immutable, however this has never been added to the specification document.

This PR closes #81, the long-standing issue to define this immutability requirement in the specification.

I hope this PR can get merged such that it becomes clear to users of this specification that mutating instances of these interfaces should not happen.


jacoscaz:
Before this gets merged we should look at modern benchmarks measuring the effects of things like `Object.freeze()` and the likes. I would personally be ok with a 5 - 10% penalty on object instantiation but anything more than that would make RDF/JS unsustainable for my use cases.

EDIT: forgot to say - I completely 100% agree with this in principle. However, fast tuple generation is critically important IMHO.

index.html
```diff
@@ -116,6 +116,9 @@ <h2>Design elements and principles</h2>
        We don't define any validation of given values (e.g. IRI, URI, CURIE). 
        Implementations that apply validation should make this fact clear in their documentation.
     </li>
+    <li>
+       Instances of <strong>Term</strong> and <strong>Quad</strong> MUST be immutable.
```
rubensworks:
```suggestion
       Instances of <strong>Term</strong> and <strong>Quad</strong> MUST be immutable, including their fields.
```

index.html
```diff
@@ -146,6 +149,7 @@ <h3><dfn>Term</dfn> interface</h3>
 
     <p>
       <a>Term</a> is an abstract interface.
+      All implementations of <a>Term</a> (including <a>Quad</a>) MUST be immutable.
```
rubensworks:
```suggestion
      All implementations of <a>Term</a> (including <a>Quad</a>) MUST be immutable, including their fields.
```

rubensworks (changes requested):

rubensworks:
> Before this gets merged we should look at modern benchmarks measuring the effects of things like Object.freeze() and the likes.

We could also say something like "... MUST be considered immutable". Then it's clear that data factories can still not use things like `freeze`, but that consumers are at fault if they mutate it.

jacoscaz:
> We could also say something like "... MUST be considered immutable". Then it's clear that data factories can still not use things like freeze, but that consumers are at fault if they mutate it.

Agreed! In the absence of benchmarks this sounds like a better way to approach this.

bergos:
> > Before this gets merged we should look at modern benchmarks measuring the effects of things like Object.freeze() and the likes.
> 
> We could also say something like "... MUST be considered immutable". Then it's clear that data factories can still not use things like `freeze`, but that consumers are at fault if they mutate it.

+1

That was also the outcome of the discussion at that time. We didn't want to force implementations to add that feature, but user should treat objects like they are immutable.

We also discussed the use case where the setters are used to change a connected persistent layer. But if one wants to implement it, they can add a note that the implementation diverges in that regard.

index.html
```diff
@@ -116,6 +116,9 @@ <h2>Design elements and principles</h2>
        We don't define any validation of given values (e.g. IRI, URI, CURIE). 
        Implementations that apply validation should make this fact clear in their documentation.
     </li>
+    <li>
+       Instances of <strong>Term</strong> and <strong>Quad</strong> MUST be immutable, including their fields.
```
smessie:
```suggestion
       Instances of <strong>Term</strong> and <strong>Quad</strong> MUST be considered immutable, including their fields.
```

index.html
```diff
@@ -146,6 +149,7 @@ <h3><dfn>Term</dfn> interface</h3>
 
     <p>
       <a>Term</a> is an abstract interface.
+      All implementations of <a>Term</a> (including <a>Quad</a>) MUST be immutable, including their fields.
```
smessie:
```suggestion
      All implementations of <a>Term</a> (including <a>Quad</a>) MUST be considered immutable, including their fields.
```

rubensworks (approved):
Looks great to me, thanks @smessie!

bergos (approved):

rubensworks:
@bergos I don't remember what the approval window duration was for PRs. Was it 2 weeks? If so, I can merge this one.

bergos:
Yes, it's two weeks. I've merged it.

awwright:
What does it mean to be "considered" immutable? This word makes the sentence intransitive, which negates its function as an interoperability requirement. At best, it suggests that programs must not modify the value, but (technically speaking) it remains mutable because this requirement may be relaxed in the future.

